PR #3, opened 2019-12-16
Move lifecycle methods to top level of export

tomjw64:


therealmarv:
you are right. Those functions are lifecycle hooks are not meant to be inside `methods` ! Well done :+1:

PR #5, opened 2019-12-16
Add style field ui and style column in backend

tomjw64:
Adds a non-autocomplete style field to the cops ui. Didn't add the autocomplete because I didn't want to make any assumptions about how we stored the autocomplete data. For now, the pipeline can use the string how it sees fit.


brittweinstein:
@tomjw64 can we merge this and move the issue over?

PR #10, opened 2020-01-06
Add locust loadtesting

therealmarv:
add simple load testing with locust


therealmarv:
Will add a Vagrant environment for this load balancer in a separate PR. Will merge now.

PR #15, opened 2020-01-08
Remove pgadmin from deploy scripts and specify backend and frontend networks

m1yag1:
Remove pgadmin from deploy scripts and specify backend and frontend networks


therealmarv:
looks good for me.
Is there a reminder (a calendar entry will do for MVP I guess) for renewing the certificate every ~2 months? Later we can automate that of course.

PR #18, opened 2020-01-09
added backend folder topic to docs and TOC

m1yag1:
* added backend folder topic to docs and TOC


c0ldcalamari:
This PR serves as an example of how to add a section to documntation for COPS

PR #25, opened 2020-01-24
added docker-compose.deploy file for rolling updates and updated deploy script

m1yag1:
* added docker-compose.deploy.settings.yml file for rolling updates. Backend and frontend now have 2 containers each. A 10 sec delay has been added before they are switched over. One container will be brought down and replaced with the updated container. 10 sec later the traffic will be routed to the new container.
* deploy script updated to use new compose file

### DIFF
--- a/docker-compose.deploy.settings.yml
+++ b/docker-compose.deploy.settings.yml
@@ -1,0 +1,18 @@
+version: "3.7"
+services:
+  backend:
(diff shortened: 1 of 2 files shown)


brittweinstein:
@helenemccarron Mike resolved the issue with Jenkins we encountered yesterday morning, could you please review this.

PR #29, opened 2020-01-30
Add basic back + front end pagination support

tomjw64:
~~WIP because I need to test backwards compatibility with the concourse pipeline. But fairly confident it will behave similarly to before.~~
Tested that the concourse resource still picks up jobs with the change to the endpoint locally using concourse 4.2.2. The only issue that could be presented is if we started >limit jobs in between the check intervals of concourse. However, this was always the case. This PR does reduce the limit to 50, though.

An important note, the `/jobs/` endpoint was kept for backwards compatibility with our concourse pipeline, but all it does now is return the first page of results, which technically is the same behaviour as before, but just reads a different way semantically.

### DIFF
--- a/backend/app/app/api/endpoints/jobs.py
+++ b/backend/app/app/api/endpoints/jobs.py
@@ -14,10 +14,25 @@ router = APIRouter()
 def list_jobs(
         db: Session = Depends(get_db),
         skip: int = 0,
-        limit: int = 100,
+        limit: int = 50,
 ):
-    """List jobs"""
-    jobs = jobs_service.get_items(db, skip=skip, limit=limit)
+    """List jobs - page 0"""
+    return list_job_page(db, 0, limit)
+
+
+@router.get("/pages/{page}", response_model=List[Job])
+def list_job_page(
+        db: Session = Depends(get_db),
+        page: int = 0,
+        limit: int = 50,
+):
+    """List jobs by page"""
+    skip = page * limit
+    order_by = jobs_service.schema_model.created_at.desc()
+    jobs = jobs_service.get_items_order_by(db,
+                                           skip=skip,
+                                           limit=limit,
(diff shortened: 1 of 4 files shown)


backend/app/app/api/endpoints/jobs.py
```diff
@@ -14,10 +14,25 @@
...
-        limit: int = 100,
+        limit: int = 50,
 ):
-    """List jobs"""
-    jobs = jobs_service.get_items(db, skip=skip, limit=limit)
+    """List jobs - page 0"""
+    return list_job_page(db, 0, limit)
+
+
+@router.get("/page/{page}", response_model=List[Job])
```
philschatz:
```suggestion
@router.get("/pages/{page}", response_model=List[Job])
```

backend/app/app/api/endpoints/jobs.py
```diff
@@ -14,10 +14,25 @@
...
+def list_job_page(
+        db: Session = Depends(get_db),
+        page: int = 0,
+        limit: int = 50,
+):
+    """List jobs by page"""
+    skip = page * limit
```
philschatz: If jobs get added while a client is paginating through the results would things get missed? It seems like a since field would work here (either by date or some other orderable field)

tomjw64: That's correct, but the risk seems extremely low. You would miss a job if all of the following occur at once:
1) the job you are looking for is nth job on a page where n==limit (1/limit chance of occurring, in this case 1/50),
2) you are searching forward in time rather than back in time (probably less than 1/2 chance of occurring, since going back will probably be far more common)
3) a new job starts between when your last page data refresh and when you go to the next page (very unlikely since we bake about 0.003 books per refresh)
4) your current page is the one right after the page you would find your job on (some variable chance)

PR #33, opened 2020-02-05
Add page limit control

tomjw64:


therealmarv:
I fixed the errors and warnings. But there is still something not right. I see this error message at the beginning always and it should not occur:

image

tomjw64:
Strange... I will look for the cause of that, but I do not recall its occurrence

therealmarv:
I think this line needs to be removed or modified to get Jobs for a page:
https://github.com/openstax/output-producer-service/blob/page-limit/frontend/store/index.js#L35

The reason for dispatching (you can also say calling getJobs) was to get the jobs on program start. Because the method is renamed/refactored this is dispatching something non existent.

PR #36, opened 2020-02-19
add Vagrantfile and shell provisioning

therealmarv:
This Vagrant environment makes sure that when you loadtest with locust you are:

- not restricted by local connection limits (like desktop mac/linux often have).
- using a DNS cache for requests so you are not load testing the DNS.

Instructions on how to use:
```
vagrant up
vagrant ssh
cd /vagrant
```

Then follow the instructions of the locust loadtesting [README](https://github.com/openstax/output-producer-service/blob/master/backend/app/tests/performance/README.md) but use http://localhost:9089 for opening the UI.

Known issues:
* This is WIP and using Python2 (wrong). You need to upgrade to Python3 in `vagrant.sh` for installation. Locust does not work with Python2 anymore.
* A docker version would be better.

### DIFF
--- a/backend/app/tests/performance/Vagrantfile
+++ b/backend/app/tests/performance/Vagrantfile
@@ -1,0 +1,14 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+# Vagrantfile API/syntax version. Don't touch unless you know what you're doing!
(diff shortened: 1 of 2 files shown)


therealmarv:
@rnathuji I assigned that PR to you. Modify it, close it... whatever works best for you.

PR #44, opened 2020-02-24
Extend load testing for COPS backend

rnathuji:
Update the Locust performance testing implementation to:

- Use Docker for underlying services
- Add load tests for GET endpoints


therealmarv:
It seems @m1yag1 recent changes add Circle CI to this repo (have not seen that before). I can change to "approve" if this is too much interlinked @rnathuji

PR #47, opened 2020-03-05
Migrate bakery pipeline tasks from cnx-recipes

rnathuji:
Migrate assemble-book, disassemble-book, and jsonify-book scripts to the bakery and integrate them into the pipeline as Concourse tasks.


bakery/src/pipeline.js
```diff
@@ -116,7 +119,10 @@ module.exports.handler = argv => {
       reportToOutputProducer(Status.PROCESSING),
       taskFetchBook(),
       taskAssembleBook(),
+      taskAssembleBookMetadata(),
       taskBakeBook(),
+      taskDisassembleBook(),
+      taskJsonifyBook(),
       taskMathifyBook(),
```
tomjw64: The fact that our content flow branches here has me thinking how this should be reflected in our pipelines. @rnathuji and cc: @philschatz do we think whenever someone builds a pdf of a book, we will want to do something with these JSON artifacts, or is work we want to do triggered by some other event?

bakery/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM python:3.7-stretch
+
+LABEL maintainer="OpenStax Content Engineering"
+
+COPY ./src/scripts /code/scripts
+COPY ./requirements.txt /code
+WORKDIR /code/
+
+RUN pip install -r requirements.txt
```
tomjw64: Since this is more of a dependency of the CLI rather than a containerization of the CLI itself, I wonder if it should live at the top level of this directory, rather than inside the scripts directory. For the same reason, I feel `output-producer-bakery` may not be the best name for it, maybe `cops-bakery-scripts`?

rnathuji: @tomjw64 - This is a good point and I debated that myself. I ended up pushing the PR with it this way, but placing the Dockerfile and `requirements.txt` in `bakery/src/scripts` was the other option I liked too. Thanks for that feedback!

Also, ack on the corresponding naming convention adjustment for the image.

bakery/src/pipeline.js
rnathuji: @tomjw64 - I'd love to discuss this more with you and @philschatz as it's 1) Something I'm unclear on myself as far as the intended interdependencies; 2) I'm still new to Concourse and don't know the best practices well, but this branching makes me wonder if we should be structuring things differently if they're fairly decoupled? For example, right now it feels wrong that if disassemble breaks mathify and beyond never happen.

rnathuji: @tomjw64 , @philschatz - Based on some discussions I had with @c0ldcalamari , I'm wondering if for now it doesn't make more sense to just merge the task definitions and leave the pipeline changes out of this PR since some of this may well be TBD and not become clear until we have more clarity from some of the other outstanding tasks (e.g. perhaps there will be a separate pipeline definition altogether for the JSON artifact generation and push of the various files to S3). Let me know if that sounds reasonable to you guys, and I'll update the PR accordingly. Conversely, if the right approach is clear at this stage, we can discuss to be sure we're all on the same page.

tomjw64: I think you're right. We should leave the pipeline changes for another day. Approved.

PR #51, opened 2020-03-18
Bn/cnx#871/extract baked book metadata

c0ldcalamari:
THE BELOW HAS BECOME OUTDATED DUE TO DECISION TO NOT INCLUDE THE DISTRIBUTION PIPELINE IN THIS PULL REQUEST.

To test:

**Start services**
```
docker-compose up
```

**Build distribution pipeline**
```
cd bakery
./build distribution-pipeline local -o dist-pipeline.local.yml
```

**Make changes in bakery/src/tasks/bake-book-metadata.js** 
uncomment `// bn16/baker-scripts`
comment out `repository: 'openstax/cops-bakery-scripts`
(because I didn't want to push the unreviewed script to cops-bakery-scripts)

**Set pipeline**, given you've already set up your local concourse/ concourse target
```
 fly -t cops-dev set-pipeline --pipeline js-pipeline --config dist-pipeline.local.yml
```

**Navigate to localhost and start a job**
collection_id: col11406
style: physics
server: katalyst01

**Notes:**

- Files get produced correctly but it doesn't get put anywhere
- We will need to change distribution-pipeline trigger to not use the output-producer-resource
- When the pipeline is configured to the distribution-pipeline it will still use the 
`output-producer-resource`
pdf-pipeline will produce a pdf-url, distribution-pipeline will not:
Screen Shot 2020-03-18 at 1 59 15 PM


bakery/README.md
```diff
@@ -29,13 +30,13 @@ and
...
-Run `./build pipeline <env> [options]...`
+Run `./build pdf-pipeline <env> [options]...`
 
 The choices for `<env>` are the basenames of the `.json` files in the `env/` directory.
 
 Examples:
-- `./build pipeline prod` -> Build the pipeline with prod environment variables and output on stdout.
-- `./build pipeline staging -o pipeline.staging.yml` -> Build the pipeline with staging environment variables and output to file `pipeline.staging.yml`, overwriting the file if it exists.
+- `./build pdf-pipeline prod` -> Build the pipeline with prod environment variables and output on stdout.
+- `./build pdf-pipeline staging -o pipeline.staging.yml` -> Build the pipeline with staging environment variables and output to file `pipeline.staging.yml`, overwriting the file if it exists.
```
rnathuji: Super minor, but should we make this `-o pdf-pipeline.staging.yml` to align with the renaming of those files?

bakery/src/tasks/build-pdf.js
```diff
@@ -24,7 +24,9 @@ const task = () => {
           dedent`
           exec 2> >(tee artifacts/stderr >&2)
           book_dir="mathified-book/$(cat book/collection_id)"
-          prince -v --output="artifacts/$(cat book/pdf_filename)" "$book_dir/collection.mathified.xhtml"
+          pdf_filename="$(cat book/collection_id)-$(cat book/version)-${'${book/server_name}'}-$(cat book/job_id).pdf"
```
rnathuji: Maybe I'm wrong, but does this work with `${book/server_name}` like this? You don't need to do something like `$(cat book/server_name)`?

bakery/src/tasks/bake-book-metadata.js
```diff
@@ -23,9 +24,14 @@ const task = () => {
...
-          mkdir "baked-book-metadata/$collection_id"
-          cp "assembled-book-metadata/$collection_id/collection.assembled-metadata.json" "baked-book-metadata/$collection_id/collection.baked-metadata.json"
-        `
+          book_dir="baked-book/$collection_id"
+          target_dir="baked-book-metadata/$collection_id"
+          mkdir "$target_dir"
+          cp "$book_dir/collection.baked.xhtml" "$target_dir/collection.baked.xhtml"
+          cp "assembled-book-metadata/$collection_id/collection.assembled-metadata.json" "$target_dir/collection.assembled-metadata.json"
+          cd "$target_dir"
+          python /code/scripts/bake-book-metadata.py collection.assembled-metadata.json collection.baked.xhtml collection.baked-metadata.json
```
rnathuji: Would it be safer, perhaps, to explicitly pass the script `"$collection_id"` just like you do for `disassemble-book.py` so we know both use the same source value as the key to put / get metadata?

bakery/src/tasks/build-pdf.js
c0ldcalamari: you're correct on this, awesome catch, thank you

bakery/src/tasks/bake-book-metadata.js
c0ldcalamari: ok, this makes sense, and in the script I can verify if the collection_id matches what is retrieved from the baked_html. 

thank you

PR #55, opened 2020-03-25
Added ui test framework to cops 

m1yag1:
Adds the ability to run ui tests against the COPS UI

This PR isn't meant to be perfect in every way. It's a good start but I'd still like to break up the integration and UI tests into a better workflow. For now, it's good feedback but we'll definitely want to make it better.

* includes page objects and region objects
* updates docker-compose files to account for new way of running tests.
* sets up pytest to accept command line arguments for running tests in chrome
* sets up registering custom markers to be used to select specific tests

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,11 +10,12 @@ jobs:
 
     steps:
       - checkout
-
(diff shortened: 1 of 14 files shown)


omehes:
This looks good to me.

PR #56, opened 2020-03-26
Add interim lookUpFeed task definition to bakery

rnathuji:
As a precursor towards defining a Distribution Pipeline, we need a
basic placeholder implementation of an ingress task that allows for
defining / triggering jobs. This change adds a simple task definition
which uses an S3 resource as input for this purpose.

### DIFF
--- a/bakery/src/tasks/look-up-feed.js
+++ b/bakery/src/tasks/look-up-feed.js
@@ -1,0 +1,34 @@
+const dedent = require('dedent')
+
+const task = () => {
+  return {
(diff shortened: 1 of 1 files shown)


rnathuji:
See notes in the parent [cnx #935](https://github.com/openstax/cnx/issues/935#issuecomment-604488819) for context / testing of these changes.

c0ldcalamari:
Success with test:
Screen Shot 2020-04-01 at 4 20 04 PM

PR #59, opened 2020-04-07
generate sha1 checksum for all resources

therealmarv:
create a resources folder, calculate checksums and define MIME type


c0ldcalamari:
Checking PR:

Ran pipeline successful

Built and ran tasks up to check sum:
```
fly -t cops-dev execute -c look-up-feed.yml -j dist-check-sum-pipeline/bakery -o book=./data/book
fly -t cops-dev execute -c fetch-book.yml -j dist-check-sum-pipeline/bakery -i book=./data/book -o fetched-book=./data/fetched-book
fly -t cops-dev execute -c assemble-book.yml -j dist-check-sum-pipeline/bakery -i book=./data/book -i fetched-book=./data/fetched-book -o assembled-book=./data/assembled-book
fly -t cops-dev execute -c assemble-book-metadata.yml -j dist-check-sum-pipeline/bakery -i book=./data/book -i assembled-book=./data/assembled-book -o assembled-book-metadata=./data/assembled-book-metadata
fly -t cops-dev execute -c bake-book.yml -j dist-check-sum-pipeline/bakery -i book=./data/book -i assembled-book=./data/assembled-book -o baked-book=./data/baked-book
fly -t cops-dev execute -c bake-book-metadata.yml -j dist-check-sum-pipeline/bakery -i book=./data/book -i assembled-book-metadata=./data/assembled-book-metadata -i baked-book=./data/baked-book -o baked-book-metadata=./data/baked-book-metadata
fly -t cops-dev execute -c checksum-book.yml -j dist-check-sum-pipeline/bakery -i book=./data/book -i baked-book=./data/baked-book -o checksum-book=./data/checksum-book
```

checksum resources successfully created in
`bakery/data/checksum-book/col26739/resources/`

therealmarv:
Input from @c0ldcalamari:

- I will rename checksum-book step to checksum-resources (or checksum-book-resources)
- More unit tests will be added

bakery/src/tasks/checksum-book.js
```diff
@@ -0,0 +1,39 @@
...
+const task = () => {
+  return {
+    task: 'checksum book',
+    config: {
+      platform: 'linux',
+      image_resource: {
+        type: 'docker-image',
+        source: {
+          repository: 'openstax/cops-bakery-scripts',
+          tag: 'dev-marvin'
```
rnathuji, line 12: Did you mean to leave this tag here?

bakery/src/scripts/checksum-resources.py
```diff
@@ -0,0 +1,137 @@
...
+
+        sha1, s3_md5 = get_checksums(img_filename)
+        mime_type = get_mime_type(img_filename)
+
+        if sha1:    # file exists
+            node.attrib['href'] = '../' + RESOURCES_DIR + '/' + sha1
+
+            create_symlink(img_filename, output_dir, sha1)
+            create_json_metadata(output_dir, sha1, mime_type, s3_md5, img_basename)
+            # print('{}, {}, {}'.format(img_filename, sha1, mime_type)) # debug
```
rnathuji, line 115: Minor cleanup: Did you want to uncomment this for logging purposes or perhaps remove?

bakery/src/tasks/checksum-book.js
therealmarv: no this tag will be removed on final merge.

bakery/src/scripts/checksum-resources.py
```diff
@@ -0,0 +1,137 @@
...
+
+        sha1, s3_md5 = get_checksums(img_filename)
+        mime_type = get_mime_type(img_filename)
+
+        if sha1:    # file exists
+            node.attrib['src'] = '../' + RESOURCES_DIR + '/' + sha1
+
+            create_symlink(img_filename, output_dir, sha1)
+            create_json_metadata(output_dir, sha1, mime_type, s3_md5, img_basename)
+            # print('{}, {}, {}'.format(img_filename, sha1, mime_type)) # debug
```
rnathuji, line 95: Minor cleanup: Did you want to uncomment this for logging purposes or perhaps remove?

bakery/src/scripts/checksum-resources.py
therealmarv: good point

bakery/src/tasks/upload-book.js
```diff
@@ -8,7 +8,8 @@ const task = ({ awsAccessKeyId, awsSecretAccessKey, bucketName }) => {
       image_resource: {
         type: 'docker-image',
         source: {
-          repository: 'openstax/cops-bakery-scripts'
+          repository: 'openstax/cops-bakery-scripts',
+          tag: 'dev-marvin'
```
rnathuji, line 12: Did you mean to leave this tag here?

bakery/src/tasks/checksum-book.js
```diff
@@ -0,0 +1,39 @@
...
+        args: [
+          '-cxe',
+          dedent`
+          exec 2> >(tee checksum-book/stderr >&2)
+          collection_id="$(cat book/collection_id)"
+          cp -r baked-book/* checksum-book
+          book_dir="checksum-book/$collection_id"
+          mkdir "$book_dir/baked"
+          find "baked-book/$collection_id/" -maxdepth 1 -type f -exec cp {} $book_dir/baked \;
+          python /code/scripts/checksum-resources.py "$book_dir"
```
rnathuji, line 31: This may tie into some work that Thomas is doing for cnx #960 around cleaning up some of our `cp -r input` operations for efficiency, but would it be possible to leave the inputs in `baked-book` and just build the resource outputs in `checksum-book` (the python script would need to take an additional argument I guess)?

I may be putting the cart before the horse here, though, so maybe worth briefly syncing with Thomas to see how perhaps (?) a minor adjustment here could align with the task changes he'll be making.

bakery/src/tasks/disassemble-book.js
```diff
@@ -13,7 +13,7 @@ const task = () => {
       },
       inputs: [
         { name: 'book' },
-        { name: 'baked-book' },
+        { name: 'checksum-book' },
```
rnathuji, line 16: This relates to my comment in `checksum-book.js`, but I think replacing this input versus adding an additional will go against the spirit of the other work to clean up the data flow in our tasks.

bakery/src/tasks/upload-book.js
```diff
@@ -38,6 +40,7 @@ const task = ({ awsAccessKeyId, awsSecretAccessKey, bucketName }) => {
           for jsonfile in "$book_dir/"*@*.json; do cp "$jsonfile" "$target_dir/$book_uuid@$book_version:$(basename $jsonfile)"; done;
           for xhtmlfile in "$book_dir/"*@*.xhtml; do cp "$xhtmlfile" "$target_dir/$book_uuid@$book_version:$(basename $xhtmlfile)"; done;
           aws s3 cp --recursive "$target_dir" s3://${bucketName}/contents
+          python /code/scripts/copy-resources-s3.py "$resources_dir" ${bucketName} resources
```
rnathuji, line 43: Just to confirm my own understanding: The main reason we are wrapping the upload in a custom script using `boto` instead of using the `aws cp / sync` CLI with `--content-type` is because the mime type value is in the JSON metadata file generated and we can't easily just upload all of a single type. Is that right?

therealmarv: That is correct. When I think about it now:
- Another approach possible is to use `aws cp` with `--content-type` but let a script generate what file lists with all jpeg, all png etc. for eatch content type.
I wonder if this speeds up the upload process. The downside with the script currently is that it is blocking on each upload which makes the upload process not the fastest. I was thinking of optimizing it with a small number of threads.
- Yet another approach is to upload everything with no content-type with `aws cp` and let the script only correct the content-type on the binary files in s3. I think that would be also faster than the current solution.

I have the feeling this requires some discussion. What do you think is the best in your opinion?

therealmarv: Update: The script has also another benefit. It checks (and verifies!) if same content already existing and does not upload it again. AFAIK `aws cp` does not have the same logic but I can be mistaken. I'm leaning of adding threads to the upload script to speed it a little bit up.

rnathuji: I don't know if it helps in the decision process, but I think this somewhat depends on how much upload control / customization capabilities we may foresee in the future for the resource files. Per your examples above, right now we get a high degree of flexibility and just need to perhaps improve the performance. On the other hand, if we know things are fairly limited in scope and the CLI saves us from having to maintain some code by using `cp --recursive` or `sync`, your ideas on ways to use that may be worth exploring. 

I'm also not sure how much we want to optimize for the double upload case (e.g. the `cp --recursive` versus `sync` in the CLI scenario) since, ideally, we never run the same distribution version twice.

tomjw64, line 43:
```suggestion
          aws s3 cp --recursive "$target_dir" "s3://${bucketName}/contents"
          python /code/scripts/copy-resources-s3.py "$resources_dir" "$bucketName" resources
```
bucketName should be quoted everywhere to prevent arg splitting.

tomjw64: @therealmarv My vote would be to use async io over threads, since this is exactly the type of work it's meant to do best :smile:

tomjw64: @therealmarv My vote would be to use async io over threads, since this is exactly the type of work it's meant to do best :smile:

bakery/src/scripts/checksum-resources.py
```diff
@@ -0,0 +1,137 @@
...
+                          namespaces={'x': 'http://www.w3.org/1999/xhtml'}):
+        img_filename = node.attrib['src']
+        img_filename = os.path.join(source_path, img_filename)
+        img_basename = os.path.basename(img_filename)
+
+        sha1, s3_md5 = get_checksums(img_filename)
+        mime_type = get_mime_type(img_filename)
+
+        if sha1:    # file exists
+            node.attrib['src'] = '../' + RESOURCES_DIR + '/' + sha1
```
tomjw64, line 91:
```suggestion
            node.attrib['src'] = '/' + RESOURCES_DIR + '/' + sha1
```
Would an absolute path work just as well?

tomjw64: Actually, I'm going to rescind this suggestion. Using a relative path allows it to work locally as well!

bakery/src/tasks/disassemble-book.js
therealmarv: I will leave the data flow like it is for now because it would be too different to the other steps.
But of course I have nothing against later optimization.

bakery/src/scripts/copy-resources-s3.py
```diff
@@ -0,0 +1,256 @@
...
+    all_resources = []
+    for sha1filename in Path(in_dir).glob(sha1_filepattern):
+        resource = {}
+        resource['bucket'] = bucket
+        resource['input_file'] = str(sha1filename)
+        resource['input_metadata_file'] = resource['input_file'] + '.json'
+        resource['output_s3'] = slash_join(
+            bucket_folder, os.path.basename(resource['input_file']))
+        resource['output_s3_metadata'] = slash_join(bucket_folder,
+                                                    os.path.basename(resource['input_file']) + '-unused.json')
```
tomjw64, line 150: seems like Path/pathlib provides good alternatives to most of the operations you are doing here if that's something that interests you. if not, it doesn't bother me.

bakery/src/scripts/copy-resources-s3.py
```diff
@@ -0,0 +1,256 @@
...
+    def s3_md5sum(s3_client, bucket_name, resource_name):
+        """ get (special) md5 of S3 resource or None when not existing """
+        try:
+            md5sum = s3_client.head_object(
+                Bucket=bucket_name,
+                Key=resource_name
+            )['ETag']
+        except botocore.exceptions.ClientError:
+            md5sum = None
+            pass
```
tomjw64, line 77: is pass necessary here?

bakery/src/tasks/checksum-book.js
```diff
@@ -0,0 +1,39 @@
+const dedent = require('dedent')
+
+const task = () => {
```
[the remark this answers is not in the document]
therealmarv: oh will add that.

bakery/src/scripts/copy-resources-s3.py
```diff
@@ -0,0 +1,256 @@
...
+            try:
+                resource = future.result()
+                if resource is not None:
+                    upload_resources.append(resource)
+                    if not disable_deep_folder_check:
+                        print('.', end='', flush=True)
+                else:
+                    if not disable_deep_folder_check:
+                        print('x', end='', flush=True)
+                check_futures.remove(future)
```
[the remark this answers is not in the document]
therealmarv: Good question. It seemed to be a best practice on many examples on the internet found.
After investigation it seemed to be because of memory consumption. After a future's result is processed it's not needed anymore in the memory and keeps the memory consumption low:
https://stackoverflow.com/questions/34770169/using-concurrent-futures-without-running-out-of-ram

Actually I will add a line after this and `del future` itself manually too like on this examples https://www.programcreek.com/python/example/86022/concurrent.futures.as_completed

bakery/src/scripts/copy-resources-s3.py
```diff
@@ -0,0 +1,256 @@
...
+            except Exception as e:
```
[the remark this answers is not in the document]
therealmarv: Getting `future.result()` can fail related to https://github.com/openstax/output-producer-service/blob/51984b42d11ff5696882b3aafa7e7c52d8f4e585/bakery/src/scripts/copy-resources-s3.py#L17-L36

It's on purpose designed this way to print out the error because of how threads fail. Normally when a thread fails you don't get the stack trace of within the thread's error message, you get a exception of the thread with no visible debug info why the thread failed. The exception is not passed into the main program and cannot be re-raised. The stack traced threads passes the error message as string to the main program and so it's visible what caused the error within the thread.
There was a lot of trial and error and failing threads on purpose manual testing involved.

bakery/src/scripts/copy-resources-s3.py
therealmarv: Yes I'm oldschool and learned about Path/pathlib too late in the process of coding this module actually (I was unfamiliar with it basically). I simply sticked to os.path because I used it already a lot in the codebase and feared of hitting corner cases with rewriting it with pathlib. I will use pathlib for the future!

bakery/src/scripts/copy-resources-s3.py
therealmarv: copy&paste error. Removed it

bakery/src/scripts/copy-resources-s3.py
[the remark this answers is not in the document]
tomjw64: Is it still possible to only wrap that single line in the try-except, since I assume we don't want a catch-all except block for the other lines?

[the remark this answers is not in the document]
therealmarv: I think I got your point now. So yes it's possible and refactored the thread processing. At the end we have two try... except blocks but with different exception handling (printing and raising). I hope this is okay. It's still necessary to step all the thread queues on both exceptions.

bakery/src/scripts/copy-resources-s3.py
```diff
@@ -0,0 +1,269 @@
...
+                halt_all_threads(executor)
+                sys.exit(1)
+            finally:
+                upload_futures.remove(future)
+                del future
+            try:
+                if result is not None:
+                    upload_count = upload_count + 1
+                    print('.', end='', flush=True)
+            except:
```
tomjw64, line 245: What can cause an exception in this try block?

bakery/src/scripts/copy-resources-s3.py
```diff
@@ -0,0 +1,269 @@
...
+                    content_type=resource['mime_type'])
+            )
+        for future in concurrent.futures.as_completed(upload_futures):
+            try:
+                result = future.result()
+            except Exception as e:
+                print(e)    # print error from ThreadPoolExecutorStackTraced
+                halt_all_threads(executor)
+                sys.exit(1)
+            finally:
```
tomjw64, line 238: Why is this finally block necessary?

bakery/src/scripts/copy-resources-s3.py
therealmarv: I think practically nothing can cause an exception here. Nevertheless if something happens here (which is very unlikely) I need to stop the threads in any case otherwise the program will not really stop entirely and will continue running even on errors here.

bakery/src/scripts/copy-resources-s3.py
```diff
@@ -0,0 +1,269 @@
...
+                upload_futures.remove(future)
+                del future
+            try:
+                if result is not None:
+                    upload_count = upload_count + 1
+                    print('.', end='', flush=True)
+            except:
+                halt_all_threads(executor)
+                raise
```
tomjw64, line 247:
```suggestion
            try:
                result = future.result()
            except Exception as e:
                print(e)    # print error from ThreadPoolExecutorStackTraced
                halt_all_threads(executor)
                sys.exit(1)
            upload_futures.remove(future)
            del future
            if result is not None:
                upload_count = upload_count + 1
                print('.', end='', flush=True)
```
This was what I was thinking. Any reason this wouldn't work?

bakery/src/scripts/copy-resources-s3.py
therealmarv: You are right, thx. I think finally is not good. I will better move the lines of the finally block into the second try block (which is also better for thread stopping... I'm pretty sure now that  threads are not stopped on errors in the final block)

bakery/src/scripts/copy-resources-s3.py
therealmarv: Normally I would agree. Ss mentioned in other comment threads will continue running this way beyond errors.

bakery/src/scripts/checksum-resources.py
```diff
@@ -0,0 +1,137 @@
...
+    """Main function"""
+    in_dir = Path(sys.argv[1]).resolve(strict=True)
+    out_dir = in_dir # overwrite baked book xhtml files
+    mkdir_resources(out_dir)
+    for xhtml_file in Path(in_dir).glob('*.xhtml'):
+        generate_checksum_resources_from_xhtml(str(xhtml_file), out_dir)
+
+
+if __name__ == "__main__":
+    main()
```
tomjw64, line 137:
```suggestion
    main()

```

PR #61, opened 2020-04-09
add baked tree json to baked metadata

c0ldcalamari:
~~Need help with, how to pythonically add a helper module correctly.~~

~~there are functions in `halp.py` that the scripts are sharing.~~

### DIFF
--- a/bakery/src/scripts/bake-book-metadata.py
+++ b/bakery/src/scripts/bake-book-metadata.py
@@ -1,28 +1,26 @@
 import sys
 import json
+import utils
 
 from lxml import etree
 from cnxepub.html_parsers import DocumentMetadataParser
+from cnxepub.collation import reconstitute
 
 raw_metadata_file, baked_xhtml_file, baked_metadata_file, collection_id = sys.argv[1:5]
-
-def parse_uri(uri):
-    if not uri.startswith('col', 0, 3): return None
-    legacy_id, legacy_version = uri.split('@')
-    return legacy_id, legacy_version
 
 def capture_book_metadata():
 
     with open(baked_xhtml_file, "r") as baked_xhtml:
         html = etree.parse(baked_xhtml)
         metadata = DocumentMetadataParser(html)
+        binder = reconstitute(baked_xhtml)
 
     required_metadata = ('title', 'license_url')
     for required_data in required_metadata:
         if getattr(metadata, required_data) is None:
             raise ValueError("A value for '{}' could not be found.".format(required_data))
         
-    legacy_id, legacy_version = parse_uri(metadata.cnx_archive_uri)
+    legacy_id, legacy_version = utils.parse_uri(metadata.cnx_archive_uri)
 
     if legacy_id != collection_id:
         print("Warning: Legacy Id '{}' does not match Collection ID '{}'.".format(legacy_id, collection_id))
@@ -30,13 +28,16 @@ def capture_book_metadata():
     with open(raw_metadata_file, "r") as raw_json :
         baked_metadata = json.load(raw_json)
     
+    tree = utils.model_to_tree(binder)
+    
     baked_book_json = {
         "title": metadata.title,
         "license": {
             "url": metadata.license_url
         },
         "legacy_id": legacy_id,
-        "legacy_version": legacy_version
+        "legacy_version": legacy_version,
+        "tree": tree
     }
     baked_metadata[collection_id] = baked_book_json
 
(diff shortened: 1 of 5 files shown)


bakery/src/scripts/disassemble-book.py
```diff
@@ -9,17 +10,6 @@
 from cnxepub.html_parsers import HTML_DOCUMENT_NAMESPACES
 from cnxepub.formatters import DocumentContentFormatter
 from cnxepub.models import flatten_to, Document, model_to_tree
```
rnathuji, line 12: Minor: Can the `model_to_tree` import be removed here with this change?

bakery/src/scripts/halp.py
```diff
@@ -0,0 +1,40 @@
...
+# (https://github.com/openstax/cnx-epub/blob/master/cnxepub/models.py#L108)
+def model_to_tree(model, title=None, lucent_id=TRANSLUCENT_BINDER_ID):
+    """Given an model, build the tree::
+        {'id': <id>|'subcol', 'title': <title>, 'contents': [<tree>, ...]}
+    """
+    id = model.ident_hash
+    if id is None and isinstance(model, TranslucentBinder):
+        id = lucent_id
+    md = model.metadata
+    title = title is not None and title or md.get('title')
```
rnathuji, line 24: Given how we're calling this in our bakery scripts where a title isn't passed, won't `title` and `clean_title` always be the same? Or is this simply because there's something downstream that's specifically expecting `clean_title`?

bakery/src/scripts/bake-book-metadata.py
```diff
@@ -1,42 +1,43 @@
...
+    
     baked_book_json = {
         "title": metadata.title,
         "license": {
             "url": metadata.license_url
         },
         "legacy_id": legacy_id,
-        "legacy_version": legacy_version
+        "legacy_version": legacy_version,
+        "tree": tree
```
rnathuji, line 40: Just a thought: Would it make sense to add some asserts to `test_bake_book()` with some of these addition metadata fields we've incorporated?

bakery/src/scripts/halp.py
c0ldcalamari: no the title won't be the same - 
```
{'legacy_id': 'col11406',
 'legacy_version': '1.17',
 'license': {'url': 'http://creativecommons.org/licenses/by/4.0/'},
 'title': 'College Physics',
 'tree': {'clean_title': 'College Physics',
          'contents': [{'clean_title': '1 Introduction: The Nature of Science '
                                       'and Physics',
                        'contents': [{'clean_title': 'Introduction to Science '
                                                     'and the Realm of '
                                                     'Physics, Physical '
                                                     'Quantities, and Units',
                                      'id': 'm42119@1.6',
                                      'slug': '1-introduction-to-science-and-the-realm-of-physics-physical-quantities-and-units',
                                      'title': '<span data-type="" itemprop="" '
                                               'class="os-text">Introduction '
                                               'to Science and the Realm of '
                                               'Physics, Physical Quantities, '
                                               'and Units</span>'},
                                     {'clean_title': 'Physics: An Introduction',
                                      'id': 'm42092@1.10',
                                      'slug': '1-1-physics-an-introduction',
                                      'title': '<span '
                                               'class="os-number">1.1</span><span '
                                               'class="os-divider"> '
                                               '</span><span data-type="" '
                                               'itemprop="" '
                                               'class="os-text">Physics: An '
                                               'Introduction</span>'}],
                        'id': '5ae762fc-302a-4170-89fd-6092688fbb58@1.17',
                        'slug': '1-introduction-the-nature-of-science-and-physics',
                        'title': '<span class="os-number">1</span><span '
                                 'class="os-divider"> </span><span '
                                 'data-type="" itemprop="" '
                                 'class="os-text">Introduction: The Nature of '
                                 'Science and Physics</span>'}],
          'id': 'col11406@1.17',
          'slug': 'college-physics',
          'title': 'College Physics'}}
```

`title` is used to generate the slug in disassemble'
'clean_title` is the title of without the html tags

Nothing downstream is expecting `clean_title`, but I thought it would be helpful but at the same time, very preemptive of me.

I'm unsure what is needed and didn't want to break the slug generation, if we use the clean title it breaks the test for disassemble script.

c0ldcalamari: I will remove clean_title hahah

rnathuji: Whoops, I completely missed the recursive part below earlier. I see / get it now .

PR #71, opened 2020-04-17
Add initial implementation of checkFeed task

rnathuji:
This is a first stab at implementing a Distribution Pipeline ingress
task which can be triggered periodically and avoids the need for a
custom resource to "pop" feed items to process. It will need to be
updated as we iron out some additional design details including:

* The current task implementation makes various assumptions around the
  feed schema which will need to be adjusted pending the outcome of
  CNX #962
* The conventions assumed for incorporating the code version in things
  like the S3 existence check will need to be adjusted pending the
  outcome of CNX #938

### DIFF
--- a/bakery/src/tasks/check-feed.js
+++ b/bakery/src/tasks/check-feed.js
@@ -1,0 +1,45 @@
+const dedent = require('dedent')
+
+const task = ({ awsAccessKeyId, awsSecretAccessKey, bucketName, feedFileUrl }) => {
+  return {
+    task: 'check feed',
+    config: {
+      platform: 'linux',
+      image_resource: {
+        type: 'docker-image',
+        source: {
+          repository: 'openstax/cops-bakery-scripts'
+        }
+      },
+      params: {
+        AWS_ACCESS_KEY_ID: `${awsAccessKeyId}`,
+        AWS_SECRET_ACCESS_KEY: `${awsSecretAccessKey}`
+      },
+      outputs: [{ name: 'book' }],
+      run: {
+        path: '/bin/bash',
+        args: [
+          '-cxe',
+          dedent`
+          exec 2> >(tee book/stderr >&2)
+          curl ${feedFileUrl} -o book-feed.json
+          feed_ids=$(cat book-feed.json | jq -r '..|.feed_id?')
+          echo "$feed_ids" | while read item
+          do
+            aws s3api head-object --bucket ${bucketName} --key "$item/.complete" || {
+              feed_entry=$(cat book-feed.json | jq -r --arg item $item '.[] | select(.feed_id==$item)')
+              echo -n "$(echo $feed_entry | jq -r '.collection_id')" >book/collection_id
+              echo -n "$(echo $feed_entry | jq -r '.server')" >book/server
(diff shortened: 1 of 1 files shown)


bakery/src/tasks/check-feed.js
```diff
@@ -0,0 +1,45 @@
...
+
+const task = ({ awsAccessKeyId, awsSecretAccessKey, bucketName, feedFileUrl }) => {
+  return {
+    task: 'check feed',
+    config: {
+      platform: 'linux',
+      image_resource: {
+        type: 'docker-image',
+        source: {
+          repository: 'openstax/cops-bakery-scripts'
```
tomjw64, line 11: Is this image used because it has aws cli installed on it? There might be a leaner image with that installation. On the other hand, this may take better advantage of Docker layer caching depending on the image `openstax/cops-bakery-scripts` is derived from. Thoughts?

bakery/src/tasks/check-feed.js
```diff
@@ -0,0 +1,45 @@
...
+        path: '/bin/bash',
+        args: [
+          '-cxe',
+          dedent`
+          exec 2> >(tee book/stderr >&2)
+          curl ${feedFileUrl} -o book-feed.json
+          feed_ids=$(cat book-feed.json | jq -r '..|.feed_id?')
+          echo "$feed_ids" | while read item
+          do
+            aws s3api head-object --bucket ${bucketName} --key "$item/.complete" || {
```
tomjw64, line 29: Will adding a `.complete` file be necessary? Is there a reason that the item's existence in s3 won't give us the same information?

bakery/src/tasks/check-feed.js
```diff
@@ -0,0 +1,45 @@
+const dedent = require('dedent')
+
+const task = ({ awsAccessKeyId, awsSecretAccessKey, bucketName, feedFileUrl }) => {
```
tomjw64, line 3: I think this makes sense, but I just want to confirm - every distribution pipeline targets it's own bucket, correct?

bakery/src/tasks/check-feed.js
rnathuji: Correct. In the past I did look around to see if there's a lean (and maintained) image with the `aws` cli but didn't find one (I used `mesosphere/aws-cli` at the time, but it's stale). That doesn't mean it wasn't a search failure on my part, though :). I also figured that since other tasks use the common image we'd benefit from image caching.

bakery/src/tasks/check-feed.js
rnathuji: Good question. This initial pass is based on the passing comment in our group discussions where we mentioned assuming some type of file marker is dropped at the end of uploads (e.g. in case an upload fails partially), but is one of those definite placeholders that I think will need to get adjusted when this is revisited in [cnx #965](https://github.com/openstax/cnx/issues/965). You can see this is part of some of the opens / questions I left there.

bakery/src/tasks/check-feed.js
rnathuji: Yes, that is indeed my working assumption, but I did add an acceptance criteria to [cnx #938](https://github.com/openstax/cnx/issues/938) where we should explicitly come to a decision on this.

PR #72, opened 2020-04-21
Add CLI test with assemble

tomjw64:
Pardon the content files. They are from business law, which is the smallest book. Used as feeder content for the tests.


tomjw64:
Now blocked by https://github.com/openstax/webview/issues/2296

.circleci/config.yml
```diff
@@ -26,6 +26,41 @@ jobs:
       - store_test_results:
           path: /tmp/test-results
 
+  cli-functional-test:
+    machine:
+      image: ubuntu-1604:201903-01
```
therealmarv, line 31: This is only minor: Can we use `ubuntu-1604:202004-01` from https://circleci.com/docs/2.0/configuration-reference/#available-machine-images to have a better match up to devs machines?

PR #75, opened 2020-04-28
Update backend test configuration to run locally

rnathuji:
This change fixes two issues:
    
 * The `tests.local.sh` script was failing due to missing arguments to
  `pytest` needed for the UI tests. That is updated to use the same
  invocation as `test.ci.sh` and the `tests-start.sh` deleted.
* The `test_jobs_post_request_successful` test was being skipped due
  to it being configured as a destructive test. We can add a setup.cfg
  file that specifies sensitive URLs so this test gets run locally and
  in CI.

### DIFF
--- a/backend/app/bin/tests-start.sh
+++ b/backend/app/bin/tests-start.sh
@@ -1,4 +1,0 @@
-#! /usr/bin/env bash
-set -e
-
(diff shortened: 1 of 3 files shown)


m1yag1:
AHH, @rnathuji I remembered that I wanted to do away with the tests-start.sh file.

PR #77, opened 2020-04-29
Support disassembled ToC linking

tomjw64:
The linking works by checking each module for a top-level id that matches
an href in the ToC. Additionally, disassembled files now have the format of
collection_id@version:page_id, since the information was needed for 
correct linking, and we formatted names like that on the upload step
anyways. i.e. the formatting takes place in disassembly now, rather than in
upload.

As an extra, this includes a change to the cops-bakery-scripts Dockerfile that takes better advantage of layer caching.

### DIFF
--- a/bakery/src/scripts/disassemble-book.py
+++ b/bakery/src/scripts/disassemble-book.py
@@ -9,7 +9,7 @@ from lxml.builder import ElementMaker, E
 from cnxepub.collation import reconstitute
 from cnxepub.html_parsers import HTML_DOCUMENT_NAMESPACES
 from cnxepub.formatters import DocumentContentFormatter
-from cnxepub.models import flatten_to, Document
+from cnxepub.models import flatten_to_documents, Document, content_to_etree
 
 def extract_slugs_from_tree(tree, data):
     """Given a tree with slugs create a flattened structure where slug data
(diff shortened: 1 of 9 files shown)


bakery/src/task.js
```diff
@@ -31,11 +31,7 @@ module.exports.builder = yargs => {
 module.exports.handler = argv => {
   const task = (() => {
     const taskFilePath = path.resolve(taskDir, `${argv.taskname}.js`)
-    try {
-      return require(taskFilePath)
-    } catch {
-      throw new Error(`Could not find task file: ${taskFilePath}`)
-    }
+    return require(taskFilePath)
```
tomjw64, line 34: This change was because I discovered this try catch did nothing but hide the real error.

bakery/src/scripts/disassemble-book.py
```diff
@@ -51,28 +51,34 @@ def main():
...
+        id_with_context = f'{collection_uuid}@{collection_version}:{doc.ident_hash}'
 
-    with open(f"{out_dir}/collection.toc-metadata.json", "w") as toc_json:
-        json.dump(book_toc_metadata, toc_json)
+        module_etree = content_to_etree(doc.content)
+        for link in nav_links:
+            link_href = link.attrib['href']
+            if not link_href.startswith('#'):
+                continue
+            if module_etree.xpath(f"/xhtml:body/xhtml:div[@id='{link_href[1:]}']", namespaces=HTML_DOCUMENT_NAMESPACES):
```
tomjw64, line 75: This assumes that the table of contents links to the `page` or `composite-page` element. Anyone making recipes would need to be aware of this (although this is how the table of contents is generated already - there is a release flag that adds composte-page links to the table of contents in this manner as well. i.e. not all books will have their links work right away - only the ones in this list https://github.com/openstax/cnx-recipes/blob/master/recipes/books/_release.scss#L232)

rnathuji:
@tomjw64 - I reviewed what's here so far and it looks good to me. I believe you're planning to add the page version removal to this PR as well so I'll hold off on approving for now. LMK if I misunderstood that conversation earlier, though.

PR #80, opened 2020-04-29
Frontend build improvements

rnathuji:


frontend/Dockerfile
```diff
@@ -11,8 +11,8 @@ RUN yarn install
 
 COPY ./ /app/
 
-ARG API_URL=https://cops.cnx.org
-ARG API_URL_BROWSER=https://cops.cnx.org
+ARG API_URL=http://backend
```
tomjw64, line 14: We just have to keep in mind that this now assumes any container created by this image must run inside a network in which the name `backend` resolves to something, which is different from before, since the name was a build arg.

PR #83, opened 2020-05-05
Add book metadata requested by REX

rnathuji:
Incorporate additional book metadata fields requested by REX and add JSON schema validation.

### DIFF
--- a/bakery/src/tasks/bake-book-metadata.js
+++ b/bakery/src/tasks/bake-book-metadata.js
@@ -40,9 +40,9 @@ const task = (taskArgs) => {
           mkdir "$target_dir"
           cp "$book_dir/collection.baked.xhtml" "$target_dir/collection.baked.xhtml"
           cat "assembled-book-metadata/$collection_id/collection.assembled-metadata.json" | jq --arg colid "$collection_id" --arg uuid "$book_uuid" --arg version "$book_version" --argjson license "$book_license" \
-              '. + {($colid): {id: $uuid, version: $version, license: $license}}' > "$target_dir/collection.metadata.json"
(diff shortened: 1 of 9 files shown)


bakery/src/tasks/bake-book-metadata.js
```diff
@@ -30,13 +31,18 @@ const task = (taskArgs) => {
...
+          book_uuid="$(cat $book_metadata | jq -r '.id')"
+          book_version="$(cat $book_metadata | jq -r '.version')"
+          book_license="$(cat $book_metadata | jq '.license')"
           book_dir="baked-book/$collection_id"
           target_dir="baked-book-metadata/$collection_id"
           mkdir "$target_dir"
           cp "$book_dir/collection.baked.xhtml" "$target_dir/collection.baked.xhtml"
-          cp "assembled-book-metadata/$collection_id/collection.assembled-metadata.json" "$target_dir/collection.assembled-metadata.json"
+          cat "assembled-book-metadata/$collection_id/collection.assembled-metadata.json" | jq --arg colid "$collection_id" --arg uuid "$book_uuid" --arg version "$book_version" --argjson license "$book_license" \
+              '. + {($colid): {id: $uuid, version: $version, license: $license}}' > "$target_dir/collection.metadata.json"
```
tomjw64, line 43: I notice this intermediate collection.metadata.json is persisted to the output (it goes to the target_dir). What exactly does this file represent? Does its fairly general name accurately represent what it is?

rnathuji: Ah, good catch. I don't think that needs to be in the output. Let me shove it into /tmp instead and call it something like `collection.baked-input-metadata.json`

PR #89, opened 2020-05-14
Add test for pdf pipeline flow and build executable and tag pinning

tomjw64:
I think this will be the final PR for issue for https://github.com/openstax/cnx/issues/958. At this point, the pipeline and CLI seem tested enough to close the issue. There can be a lot more confidence in charges with these tests in CI.

### DIFF
--- a/bakery/src/tests/test_cli.js
+++ b/bakery/src/tests/test_cli.js
@@ -1,13 +1,24 @@
 const test = require('ava')
 const { spawn } = require('child_process')
 const fs = require('fs-extra')
+const path = require('path')
+const yaml = require('js-yaml')
+const dedent = require('dedent')
 
 const completion = subprocess => {
   const error = new Error()
   return new Promise((resolve, reject) => {
+    let stdout = ''
+    let stderr = ''
+    subprocess.stdout.on('data', data => {
+      stdout += data.toString()
+    })
+    subprocess.stderr.on('data', data => {
+      stderr += data.toString()
+    })
     subprocess.on('exit', code => {
       if (code === 0) {
-        resolve(undefined)
+        resolve({ stdout, stderr })
       } else {
         error.message = `Subprocess failed with code ${code}`
         reject(error)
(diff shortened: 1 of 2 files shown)


bakery/src/tests/test_cli.js
```diff
@@ -16,6 +27,75 @@ const completion = subprocess => {
...
+  const buildPipelineResult = await completion(buildPipeline)
+
+  const obj = yaml.safeLoad(pipelineOut)
+  const sources = sourceObjs(obj)
+  for (const source of sources) {
+    t.is(source.tag, customTag, formatSubprocessOutput(buildPipelineResult))
+  }
+})
+
 test('stable flow in content distribution pipeline', async t => {
```
rnathuji, line 99: Minor: Update the test label text perhaps (since it now runs both dist and PDF tests)?

bakery/src/tests/test_cli.js
```diff
@@ -26,82 +106,108 @@ test('stable flow in content distribution pipeline', async t => {
...
     'assemble',
     bookId
-  ], { stdio: 'inherit' })
-  await completion(assemble)
-  t.truthy(fs.existsSync(`${outputDir}/${bookId}/assembled-book/${bookId}/collection.assembled.xhtml`))
+  ])
+  const assembleResult = await completion(assemble)
+  t.truthy(fs.existsSync(`${outputDir}/${bookId}/assembled-book/${bookId}/collection.assembled.xhtml`), formatSubprocessOutput(assembleResult))
 
   const assembleMeta = spawn('node', [
```
rnathuji, line 132: Minor: I believe this task is also only for distribution, so maybe it can be used to fork the dist branch instead of `bakeMeta` if you're trying to keep everything encapsulated in `branchDistribution`?  I may be forgetting some interdependencies there, though, and this is a bit more logical in terms of ordering flow, so ... :shrug:

tomjw64: I decided to branch them where they ceased to have any common steps, in this case, both still need to bake, so I saw no reason to branch here. It was more about taking advantage of already completed steps than logical separation to keep CI times dow (clearly, as I used one test to cover both pipelines)

PR #91, opened 2020-05-15
Use tagged cnx-recipes-output image over git repository

tomjw64:
This is a dependency of locking the code versions for pipelines

### DIFF
--- a/bakery/src/pipelines/distribution.js
+++ b/bakery/src/pipelines/distribution.js
@@ -16,10 +16,11 @@ const pipeline = (env) => {
 
   const resources = [
     {
-      name: 'cnx-recipes',
-      type: 'git',
+      name: 'cnx-recipes-output',
+      type: 'docker-image',
       source: {
-        uri: 'https://github.com/openstax/cnx-recipes.git'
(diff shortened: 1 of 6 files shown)


bakery/src/pipelines/distribution.js
```diff
@@ -16,10 +16,11 @@ const pipeline = (env) => {
...
   const resources = [
     {
-      name: 'cnx-recipes',
-      type: 'git',
+      name: 'cnx-recipes-output',
+      type: 'docker-image',
       source: {
-        uri: 'https://github.com/openstax/cnx-recipes.git'
+        repository: 'openstax/cnx-recipes-output',
+        tag: env.IMAGE_TAG || 'latest'
```
rnathuji, line 23: I noticed there's not currently a `latest` tag of this image pushed up yet.

tomjw64: Wow, good catch.

tomjw64: I pushed a latest image for cnx-recipes-output - just a copy of `20200514.135335` for now

PR #95, opened 2020-05-18
Update distribution pipeline to use a prefix on uploads

rnathuji:
This PR includes the change to incorporate a prefix for our distribution S3 uploads as well as a few minor improvements including:

* Fix a minor issue in `look-up-feed` so stderr is placed in the task output
* Pass `versionedFile` value used in the S3 resource as an input to the `look-up-feed` task so they are in sync
* Make minor usage / naming changes to the PDF and distribution S3 bucket name environment variables

### DIFF
--- a/bakery/src/tasks/look-up-feed.js
+++ b/bakery/src/tasks/look-up-feed.js
@@ -24,7 +24,7 @@ const task = (taskArgs) => {
         args: [
           '-cxe',
           dedent`
-          exec 2> >(tee s3-feed/stderr >&2)
+          exec 2> >(tee book/stderr >&2)
           feed=s3-feed/distribution-feed.json
           echo -n "$(cat $feed | jq -r '.[-1].collection_id')" >book/collection_id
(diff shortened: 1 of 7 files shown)


bakery/src/tasks/look-up-feed.js
```diff
@@ -3,6 +3,7 @@ const dedent = require('dedent')
 const { constructImageSource } = require('../task-util/task-util')
 
 const task = (taskArgs) => {
+  const { versionedFile } = taskArgs
```
tomjw64, line 6: I attempted to design all tasks to not throw when they receive `undefined` as a parameter. Destructuring a null/undefined will throw. Might be good to keep with the tradition. Could either do something like the following or pair it into one block with line 10.
```suggestion
  const { versionedFile } = taskArgs == null ? {} : taskArgs
```

tomjw64: Actually, I'll rescind this. It's a fairly arbitrary requirement and it looks like this task will probably fail anyways if it's not given.

PR #96, opened 2020-05-20
Remove use of neb config file in fetch-book

rnathuji:
Recent improvements to `neb` allow users to invoke commands without
creating a configuration file. Specifically, this change makes use
of the fact that with those updates, if you pass a fully qualified
hostname to neb as the environment and no configuration is available
that matches, it will use the environment value as the cnx hostname.


rnathuji:
Opening this as draft PR for the time being as I want to test this change in a local pipeline once the corresponding `neb` changes are merged and a corresponding tag is available.

PR #97, opened 2020-05-20
Bn/cnx#929/production ready details

c0ldcalamari:
Branch for documentation of COPS - 

```
docker-compose up -d
```

navigate to localhost:8000 to review documentation

Documentation should reflect good bones so we can continue to add to it.

----

Quick before and after:
Before:
Screen Shot 2020-06-08 at 9 36 54 AM

After:
Screen Shot 2020-06-08 at 9 37 21 AM


bakery/src/pipeline.js
```diff
@@ -10,22 +10,22 @@ module.exports.command = commandUsage
...
-  yargs.usage(`Usage: ${process.env.CALLER || 'build.js'} ${commandUsage}`)	
-  yargs.positional('pipelinetype', {	
-    describe: 'type of pipeline',	
-    choices: fs.readdirSync(pipelineDir).map(file => path.basename(file, '.js')),	
-    type: 'string'	
-  }).positional('env', {
-    describe: 'name of environment',
+  yargs.usage(`Usage: ${process.env.CALLER || 'build.js'} ${commandUsage}`)
+  yargs.positional('env', {
+    describe: 'pipeline environment',
```
rnathuji, line 15: Was the goal of the changes to this file to improve the documentation of the positional arguments on the command, or something functional? (e.g. is it just trying to change the `describe` lines versus the reordering of the args themselves being intentional)

bakery/src/task.js
```diff
@@ -9,7 +9,7 @@ module.exports.command = commandUsage
 module.exports.aliases = ['t']
 module.exports.describe = 'builds a bakery pipeline task runnable with fly execute'
 module.exports.builder = yargs => {
-  yargs.usage(`Usage: ${process.env.CALLER || 'build.js'} ${commandUsage}`)
+  yargs.usage(`Usage: ${process.env.CALLER || 'build'} ${commandUsage}`)
```
rnathuji, line 12: Just curious: I believe this to handle where `process.env.CALLER` doesn't have a value and we want to display `build` vs `build.js` in the output. If so, how are you invoking it to get the `||` path?

bakery/src/pipeline.js
c0ldcalamari: reordering the args themselves, if i remember correctly.

c0ldcalamari: will confirm - what are your thoughts?

docs/.python-version
```diff
@@ -0,0 +1 @@
+cnx-automation
```
rnathuji, line 1: This looks like perhaps an accidental file being added to the PR (?)

bakery/src/pipeline.js
rnathuji: I don't think it does reorder them (?) -- at least based on a quick test, but could be user error on my part =). I think to do that you'd need to change line 7 (I'm not super familiar with `yargs`, though). 

However, that being said, I'm not sure the order of `<env> <type>` makes more sense than the current `<type> <env>` (at least in my mind 🤷 )

c0ldcalamari: I will double check on this all, thank you!

docs/README.md
```diff
@@ -0,0 +1,28 @@
+## Quick Start to Docs
```
rnathuji, line 1: This documentation seems somewhat redundant relative to [the main README.md section](https://github.com/openstax/output-producer-service/tree/bn/cnx%23929/production-ready-details#view-the-docs). Should the improvements be made there, or perhaps that be moved here?

Also, there's a note there about documentation automatically rebuilding. If this isn't working as intended, perhaps there's a bug? (I see we mount files as a volume and also copy files in the `Dockerfile` -- maybe that's causing an issue).

docs/distribution/cli.rst
```diff
@@ -0,0 +1,8 @@
+.. _distribution-pipeline-cli:
```
rnathuji, line 1: Minor organizational comment: Since the CLI is for both PDF and Distribution tasks, is there a better place to have this file live in terms of TOC so users don't think it only applies to the latter?

docs/operations/cleaning_up_the_swarm.rst
```diff
@@ -0,0 +1,133 @@
...
+
+##################
+Clean Up The Swarm
+##################
+
+Docker swarm does not come with any "garbage collection" for dangling 
+volumes or unused containers that have been created during updates or after 
+restarts.
+
+**To Do: Add Cronjob to run** ``docker prune``
```
rnathuji, line 11: Minor: I think we can remove this TODO as the ansible playbook we added creates this cron job

docs/operations/generate_pipeline_config.rst
```diff
@@ -0,0 +1,126 @@
...
+
+1. `Install Docker <https://docs.docker.com/get-docker/>`_
+==========================================================
+
+2. Clone copy of output-producer-service
+========================================
+
+.. code-block:: bash
+
+    $ git cllone git@github.com:openstax/output-producer-service.git
```
rnathuji, line 19: Minor: Noticed typo on `cllone` --> `clone` here.

docs/operations/updating_the_stack.rst
```diff
@@ -1,133 +1,200 @@
 .. _operations-updating-the-stack:
```
rnathuji, line 2: Just FYI, I there are some doc changes in [this recent PR](https://github.com/openstax/output-producer-service/pull/106/files) which may need to get massaged in with your updates to this page.

docs/README.md
c0ldcalamari: To note: The documentation does rebuild for the most part, but sometimes the navigation changes don't take when updating the index.rst/ the toc tree.

m1yag1: You have to restart the entire container if you change navigational pieces.

bakery/src/task.js
c0ldcalamari: I have not been able to invoke it to get the `||` path. But it looks like I may have been confused because `build.js` doesn't do anything..?

```bash
(cops) [Tue Jun 16 15:15:01]brenda@brenda-openstax:bakery[bn/cnx#929/production-ready-details !?] $ build.js task --help
-bash: build.js: command not found
(cops) error 127 [Tue Jun 16 15:16:46]brenda@brenda-openstax:bakery[bn/cnx#929/production-ready-details !?] $ ./build task --help
Usage: ./build task <taskname> [options]...
...
```

Where does the display of `build` vs `build.js` happen?

c0ldcalamari: I put `.js` back. I don't clearly remember why I removed it, but I still don't understand it.

c0ldcalamari:
To @ripal's notes on the `.js` files to build pipeline and tasks. I checked out the master version of `pipeline` and read over the `yarg` commands and used them as described. Everything looks fine/ works.

The only thing that is odd is that under `Command` `yargs` is advising to use `build.js`

`build.js pipeline <pipelinetype> <env> `

This is most likely because `build.js` is requiring command from `pipeline.js` / `task.js` but don't really know how to get rid of the `build.js` part and I think this is incorrect.

bakery/src/pipeline.js
c0ldcalamari: I put things back, like how they are on master now.

PR #100, opened 2020-05-20
Correctly put server shortname in pdf url, add test

tomjw64:


bakery/src/tasks/look-up-book.js
```diff
@@ -32,11 +32,11 @@ const task = (taskArgs) => {
           cp output-producer/version book/version
           cp output-producer/collection_style book/style
           cp output-producer/content_server book/server
-          pdf_filename="$(cat book/collection_id)-$(cat book/version)-$(cat book/server_name)-$(cat book/job_id).pdf"
           echo "$pdf_filename" > book/pdf_filename
           wget -q -O jq 'https://github.com/stedolan/jq/releases/download/jq-1.6/jq-linux64' && chmod +x jq
-          server_name="$(cat output-producer/job.json | ./jq -r '.content_server.name')"
-          echo "$server_name" >book/server_name
+          server_shortname="$(cat output-producer/job.json | ./jq -r '.content_server.name')"
+          echo "$server_shortname" >book/server_shortname
```
rnathuji, line 38: I'm not sure `book/server_shortname` is used anywhere outside of this task. Is it just a convenience in case we want / need it later on?

bakery/src/tasks/look-up-book.js
```diff
@@ -32,11 +32,11 @@ const task = (taskArgs) => {
...
+          pdf_filename="$(cat book/collection_id)-$(cat book/version)-$(cat book/server_shortname)-$(cat book/job_id).pdf"
```
rnathuji, line 39: Should the line `echo "$pdf_filename" > book/pdf_filename` also be moved after this to write it out?

bakery/src/tests/test_cli.js
```diff
@@ -196,6 +196,7 @@ test('stable flow in pdf and distribution pipeline', async t => {
     ])
     const buildPdfResult = await completion(buildPdf)
     t.truthy(fs.existsSync(`${outputDir}/${bookId}/artifacts/collection.pdf`), formatSubprocessOutput(buildPdfResult))
+    t.is(fs.readFileSync(`${outputDir}/${bookId}/artifacts/pdf_url`, { encoding: 'utf8' }), 'https://none.s3.amazonaws.com/collection.pdf', formatSubprocessOutput(buildPdfResult))
```
rnathuji, line 199: I may well be confused at this point, but is this the expected filename? (the test passes,  know, I'm just wondering if it shouldn't be checking for something of the form `-version-server_shortname-job_id.pdf`)

bakery/src/tasks/look-up-book.js
tomjw64: yeah, I thought i'd leave it in place for the time being

bakery/src/tasks/look-up-book.js
tomjw64: Good catch

m1yag1: makes me think that perhaps it's best to build that filename via a function with arguments defined? seems like an easy thing to mess up so why not have it be generated?

bakery/src/tests/test_cli.js
tomjw64: there are limits to what this can test, because the lookup step does not actually run. It's mainly ensuring that nothing erroneous is added beyond the given filename. It would have at least caught the extra ')' that we were seeing.

rnathuji: Ah, that's right -- The CLI is going to always give it `collection.pdf` as `pdf_filename`. 

Couldn't think of a way to give better coverage given the limitations.

bakery/src/tasks/look-up-book.js
rnathuji: Hmmm...I'm not sure I understand. It's computed in one place here and saved for future reference. So, in effect, this is a function execution for other tasks. @m1yag1 is your suggestion to add a function def in the task bash or something?

PR #101, opened 2020-05-21
CLI sync w/ direct download as a fallback

tomjw64:
This should help people who encounter problems syncing from concourse v4 to concourse v6.

Installing fly is also no longer a requirement to use the cli, as it will be able to download it from the concourse container that gets spun up now and stored in `$HOME/.local/bin/bakery-cli-fly`

### DIFF
--- a/bakery/src/cli/execute.js
+++ b/bakery/src/cli/execute.js
@@ -126,22 +126,40 @@ const flyExecute = async (cmdArgs, { image, persist }) => {
       output: 'silent'
     })
 
+    console.log('syncing fly')
+    let flyPath
     try {
-      console.log('syncing')
-      const sync = spawn('fly', [
-        'sync',
-        '-c', 'http://localhost:8080'
-      ], { stdio: 'inherit' })
-      children.push(sync)
-      await completion(sync)
-    } catch (err) {
-      console.log('current fly cli incompatible with this concourse version')
-      console.log('syncing fly cli via direct download')
-      const printOldFlyVersion = spawn('fly', ['--version'], { stdio: 'inherit' })
+      flyPath = which.sync('fly')
+    } catch {
+      console.log('no fly installation detected on PATH')
+      const flyDir = path.resolve(process.env.HOME, '.local/bin/')
+      flyPath = path.resolve(flyDir, 'bakery-cli-fly')
+      fs.mkdirSync(flyDir, { recursive: true })
+    }
+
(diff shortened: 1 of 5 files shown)


bakery/src/cli/execute.js
```diff
@@ -125,13 +126,36 @@ const flyExecute = async (cmdArgs, { image, persist }) => {
...
-    await completion(sync)
+    try {
+      console.log('syncing')
+      const sync = spawn('fly', [
+        'sync',
+        '-c', 'http://localhost:8080'
+      ], { stdio: 'inherit' })
+      children.push(sync)
+      await completion(sync)
+    } catch (err) {
```
tomjw64, line 137: What if user does NOT have fly installed despite the installation instructions to do so. Should we just download the correct fly version from the concourse container and use it anyways?

bakery/src/cli/execute.js
```diff
@@ -125,13 +126,36 @@ const flyExecute = async (cmdArgs, { image, persist }) => {
...
+      console.log('current fly cli incompatible with this concourse version')
+      console.log('syncing fly cli via direct download')
+      const printOldFlyVersion = spawn('fly', ['--version'])
```
rnathuji, line 140: This works for me as far as the manual sync (e.g. when I run with fly 5.8 I end up with fly 6.1), but I didn't see the two version outputs to the screen which I think this is for?:
```
syncing
error: unknown flag `c'
current fly cli incompatible with this concourse version
syncing fly cli via direct download
logging in
logging in to team 'main'
```

bakery/src/cli/execute.js
rnathuji: I think this would be a nice user experience. Without `which` resolving would you just drop it in the current directly and add that to the path for the CLI?

bakery/src/cli/execute.js
tomjw64: Good eye. I need to add `{ stdio: inherit }`

bakery/src/cli/execute.js
rnathuji: I like the idea of using `.local/bin` in `$HOME`

tomjw64: Now uses `$HOME/.local/bin/bakery-cli-fly` as a last resort, but will prefer `fly` on the path and sync it if needed.

PR #103, opened 2020-05-26
Add page metadata requested by REX

rnathuji:


bakery/src/scripts/assemble-book-metadata.py
```diff
@@ -1,18 +1,48 @@
...
+    with open(input_assembled_file, "r") as in_file:
+        binder = reconstitute(in_file)
+
+    for doc in flatten_to_documents(binder):
+        # Parse the metadata from the supporting module XHTML files generated
+        # by neb during assemble to gain access to metadata that is in the
+        # head and doesn't get incorporated into the colllection assembled
+        # XHTML
+        module_xhtml_metadata = get_module_xhtml_metadata(
+            in_dir / f"{doc.id}.xhtml"
```
tomjw64, line 36: This seems to be examining individual modules. Two things seem odd to me:
1. This seems like this must wait for assembly to finish.
2. It relies on the assembled-book input, which is not the source of these files

I'm not sure this is a blocker, but I thought I would pick your brain about those.

tomjw64: woops, didn't send this in in time!

rnathuji: It's never too late =). Happy to chat about the approach here to see if there's a better way, but probably easier to walk through and talk live. Want to hop on a call to discuss later this AM?

As far as the question for input, yes, it does assume assembly, but I'm not sure I understand the comment on the [assembled-book input](https://github.com/openstax/output-producer-service/blob/fa7d2af7ede80f71486cf53d8425e8fff9b629e8/bakery/src/tasks/assemble-book-metadata.js#L22) not being the source of the files.

PR #104, opened 2020-05-27
Add stop + up commands

tomjw64:
`stop` will help clean up any containers that are left running for whatever reason, and there is a `-d` option to destroy the containers. `up` will start the containers again, but is probably only useful for debugging purposes.

Because the requirements for these kinds of commands are very different, task commands have been scoped under the `run` subcommand. E.g. `fetch` -> `run fetch`. Documentation has been changed to reflect this.

The restructuring should also help prepare for removal of the `-c` flag and packaging for npm.

### DIFF
--- a/bakery/src/cli/execute.js
+++ b/bakery/src/cli/execute.js
@@ -186,7 +186,7 @@ const flyExecute = async (cmdArgs, { image, persist }) => {
 }
 
 const tasks = {
-  fetch: () => {
+  fetch: (parentCommand) => {
     const commandUsage = 'fetch <server> <collid> <version>'
     const handler = async argv => {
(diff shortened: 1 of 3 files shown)


bakery/src/cli/execute.js
```diff
@@ -499,8 +495,8 @@ const yargs = require('yargs')
         handler(argv).catch((err) => { console.error(err) })
       }
     }
-  }).call())
-  .command((() => {
+  },
+  checksum: (parentCommand) => {
```
rnathuji, line 499: Minor (and not directly related to this PR): Is the comment below on line 500 intended to be there, or can it be cleaned up?

PR #109, opened 2020-06-01
additional cops tests - basic cops modal tests

omehes:


brittweinstein:
@omehes please provide descriptions of the tests you've added.

omehes:
Added 2 basic tests to verify cops modal:
1. inputting data and clicking cancel - modal closes
2. no data input and clicking create button - modal remain open

backend/app/tests/ui/pages/home.py
```diff
@@ -2,13 +2,42 @@
...
+    )
+    _go_button_locator = (
+        By.CSS_SELECTOR,
+        "#app > div.v-application--wrap > main > div > div > div "
+        "> div > div.d-md-flex > button:nth-child(5) > span",
+    )
+    _page_field_locator = (By.CSS_SELECTOR, "#input-17")
+    _no_data_available = (
+        By.CSS_SELECTOR,
+        "# app > div > main > div > div > div > div > "
```
m1yag1, line 38: Will this css selector work? I see a space between `#` and `app`.

backend/app/tests/ui/pages/home.py
```diff
@@ -23,16 +52,74 @@ def is_create_new_pdf_button_displayed(self):
...
 
         def click_cancel_button(self):
             self.find_element(*self._modal_cancel_button_locator).click()
             return self
+
+        def click_create_button(self):
+            self.find_element(*self._modal_create_button_locator).click()
+            return self
+
+        def fill_collection_id_field(self, value):
```
m1yag1, line 111: I like the use of the TAB key here (lots of potential for accessibility testing) but it's a bit misleading, no? Let's say I'm writing a test and for some reason I want to jump ahead to `fill_version_field` and the focus of the element is not the one previous to it. The name of the method gives me the impression I'll be able to fill the version field but actually all I'm doing is hitting tab and entering a value. In other words, I will not actually be filling the version field.

I think it would be better to change this code to select the proper input box and then send the value rather than tabbing through. I do thing tabbing through could be a separate test that compares you do things in the right order for the future.

backend/app/tests/ui/test_cops_modal.py
```diff
@@ -0,0 +1,51 @@
...
+    modal.click_create_button()
+
+    # THEN: The create new pdf modal stays open
+    assert home.create_pdf_modal_is_open
+
+
+@pytest.mark.smoke
+@pytest.mark.ui
+@pytest.mark.nondestructive
+@pytest.mark.parametrize("colid", ["col26069"])
```
m1yag1, line 27: these can be combined into one parametrize. 
Example: 
https://github.com/openstax/cnx-automation/blob/master/tests/neb/ui/test_get.py#L79

Documentation
https://docs.pytest.org/en/latest/parametrize.html#pytest-mark-parametrize

omehes:
@m1yag1 As per your PR review, I changed parametrize and css locator and also reworked the page objects for filling in the data field in cops modal. However, I am having issues with send_keys() attribute ("Unresolved attribute reference 'send_keys' for class 'CreatePDFModal'") and per our conversation, here it is for you to see and try for possible issues.

m1yag1:
Closing PR. Will be reopened as a Draft PR.

PR #113, opened 2020-06-08
tests modal fields errors when no input happens and create button is clicked

omehes:
as per our last week's discussion, I created this test to verify the error messages of each input field in the cops modal. I added temp css_selectors until Marvin gets new ones (as per your comment from last Friday).


omehes:
As per Mike's and Marvin's comments and help, added locators to index.vue and changed locastors in home.py

frontend/pages/index.vue
```diff
@@ -41,16 +41,21 @@
                     <v-col cols="12" sm="3" md="3">
                       <v-text-field
                         v-model="collectionId"
-                        :rules="collectionRules"
+                        :rules="[v => !!v || 'Collection ID is required']"
+                        :items="collectionRules"
```
therealmarv, line 45: rules should not replaced with items.
The rules for checking collection ID are checking two things:
the rule in line 44 and that collection is expressed in a `col` + number format.
See lines https://github.com/openstax/output-producer-service/blob/15acaca03d7da6a0bb3074e4d91deb252170c7f6/frontend/pages/index.vue#L221-L224

frontend/pages/index.vue
```diff
@@ -41,16 +41,21 @@
...
+                        class="collection-id-error-text"
                         hint="e.g. col12345"
                         required
                       />
                     </v-col>
                     <v-col cols="12" sm="3" md="3">
                       <v-text-field
                         v-model="version"
+                        :rules="[v => !!v || '']"
+                        :items="versionItems"
```
therealmarv, line 56: versionItems should not be added because we don't have a list of versions to choose from currently.

If rules are added here then it's not allowed to have an empty version box. My assumption was that empty version editbox=latest version. If that should be changed then an error text should be added instead of `''`

omehes: @therealmarv I removed the added lines for version. I will only test colid, style and server error messages.

frontend/pages/index.vue
omehes: @therealmarv All the fields have this construction
```
:rules="[v => !!v || 'Style is required']"
                        :items="styleItems"
```
I just copied them to each modasl field. Should I change them to the one you showed above? 
What is the .test in ```/^col\d*$/.test(v)```

therealmarv: I think for this one remove the new added `:items` completely and leave line 44 the same as before:
```
:rules="collectionRules"
```

your test should still work. If you have any more question just ping me somewhere in Slack.

omehes:
@therealmarv I changed index.vue and other as per your comments (removed changes made for testing version).

frontend/pages/index.vue
```diff
@@ -41,8 +41,9 @@
                     <v-col cols="12" sm="3" md="3">
                       <v-text-field
                         v-model="collectionId"
-                        :rules="collectionRules"
+                        :rules="[v => !!v || 'Collection ID is required']"
```
therealmarv, line 44: please replace with previous line:
```
:rules="collectionRules"
```

omehes: I misunderstood... :) done

PR #115, opened 2020-06-09
Update bakery scripts to reflect neb changes for deterministic UUIDs

rnathuji:


rnathuji:
Note: The `cli-functional-test` is currently failing due to corresponding [changes in neb](https://github.com/openstax/nebuchadnezzar/pull/182) not being merged yet.

bakery/src/scripts/disassemble-book.py
```diff
@@ -46,15 +46,11 @@ def main():
...
-        # It's important that we generate slug metadata in parallel with disassemble
-        # so that where ident_hash values are based upon potentially randomly
-        # generated UUIDs we can still use them as unique keys in JSON outputs
-        # without diverging
         slugs = extract_slugs_from_binder(binder)
 
     with open(baked_metdata_file, "r") as baked_json:
         baked_metadata = json.load(baked_json)
-        book_toc_metadata = baked_metadata.get(binder.id)
+        book_toc_metadata = baked_metadata.get(binder.ident_hash)
```
tomjw64, line 53: Since the binder returns ident with uuid now, does the script still need to take a collection_uuid + collection_version argument?

rnathuji: Technically, no, but similar to bake-book where the assumption around the composition of the ident_hash required a change this time around and I explicitly pulled the values by name from the `metadata.json` instead, I thought keeping those inputs explicit would be safer. Maybe that's too paranoid?

rnathuji: I think I accidentally deleted my previous response when I went to edit so this may be a repeat (sorry if so): Technically, no, but I thought it would be safer to have those passed explicitly instead of parsing from the `ident_hash` to avoid having an inferred expectation of how that value is constructed in the code. Am I overthinking it?

rnathuji:
The `cli-functional-test` is now passing with the merged version of neb (v9.8.0)

PR #116, opened 2020-06-15
Add poetry for dependency management

m1yag1:
I noticed that there were several alerts from GitHub on this repo about out of date dependencies. There have been some conversations going on about python packaging and the absence of a .lock file that is similar to other languages like ruby and javascript. I've added `poetry` into the project which has this functionality and also allows us to use it to do upgrades of our dependencies. I'm wondering what other devs think about this so am submitting this PR for feedback.

To aid in review I recommend devs look at what has changed in the `backend.dockerfile` and look at the two files required by poetry -- the pyproject.toml and poetry.lock file. Within the Dockerfile you'll see that I used multi-staged builds that are later used as "targets" in the docker-stack.yml for what build/runner stages to use. This makes it way easier to maintain different images for production or development. This should also help with docker image sizes as this tells docker when it can throw specific "layers" away from the built up image.


m1yag1:
This is the difference between the multi-stage vs non multi-stage builds:
image

I expect production to be even smaller b/c it doesn't require as many dev dependencies.

rnathuji:
Overall `poetry` seems pretty interesting and straight forward to incorporate. Is the only downside having to adopt an additional tool, or am I missing something?

backend/backend.dockerfile
```diff
@@ -1,29 +1,61 @@
...
+
+# install poetry
+# keep this in sync with the version in pyproject.toml and Dockerfile
+ENV POETRY_VERSION 1.0.5
+RUN curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py | python
+ENV PATH "/root/.poetry/bin:/opt/venv/bin:${PATH}"
+
+# install wait-for-it
+# allows us to wait for specific services to be up before executing scripts
 RUN curl https://raw.githubusercontent.com/vishnubob/wait-for-it/54d1f0bfeb6557adf8a3204455389d0901652242/wait-for-it.sh \
```
tomjw64, line 20: wait-for-it doesn't seem to do anything within the lifetime of the `builder` stage. Should it live in the `runner` stage instead?

backend/backend.dockerfile
```diff
@@ -1,29 +1,61 @@
-FROM python:3.7
+FROM python:3.8-slim as base
 
-LABEL maintainer="OpenStax Content Engineering"
+RUN apt-get update -qq \
+ && apt-get install -y --no-install-recommends \
+    # required by psycopg2 at build and runtime
+    libpq-dev \
+    curl \
```
tomjw64, line 7: is `curl` also required during runtime or just `libpq-dev`?

backend/app/pyproject.toml
```diff
@@ -0,0 +1,85 @@
...
+jinja2 = "^2.11.2"
+mako = "^1.1.3"
+markupsafe = "^1.1.1"
+psycopg2-binary = "^2.8.5"
+pydantic = "^1.5.1"
+starlette = "0.13.2"
+tenacity = "^6.2.0"
+uvicorn = "^0.11.5"
+
+[tool.poetry.dev-dependencies]
```
tomjw64, line 25: Do these dependency lists include things that are not top-level dependencies, and do we want to keep it that way? I feel like we really only get the full benefit of poetry if we minimize the practice of declaring the versions for dependencies of dependencies.

m1yag1: That's a good point. Let me do some more testing and see if I can get it to only install top level dependencies. I had only fastapi and starlette above and i was getting errors that it didn't have uvicorn. This gave me the impression that poetry installs ONLY what you tell it.

tomjw64:
```
$ poetry show -t --no-dev
alembic 1.4.2 A database migration tool for SQLAlchemy.
├── mako *
│   └── markupsafe >=0.9.2 
├── python-dateutil *
│   └── six >=1.5 
├── python-editor >=0.3
└── sqlalchemy >=1.1.0
click 7.1.2 Composable command line interface toolkit
fastapi 0.56.0 FastAPI framework, high performance, easy to learn, fast to code, ready for production
├── pydantic >=0.32.2,<2.0.0
└── starlette 0.13.2
gunicorn 20.0.4 WSGI HTTP Server for UNIX
└── setuptools >=3.0
h11 0.9.0 A pure-Python, bring-your-own-I/O implementation of HTTP/1.1
httptools 0.1.1 A collection of framework independent HTTP protocol utils.
jinja2 2.11.2 A very fast and expressive template engine.
└── markupsafe >=0.23
mako 1.1.3 A super-fast templating language that borrows the  best ideas from the existing templating languages.
└── markupsafe >=0.9.2
markupsafe 1.1.1 Safely add untrusted strings to HTML/XML markup.
psycopg2-binary 2.8.5 psycopg2 - Python-PostgreSQL Database Adapter
pydantic 1.5.1 Data validation and settings management using python 3.6 type hinting
python-dateutil 2.8.1 Extensions to the standard Python datetime module
└── six >=1.5
six 1.15.0 Python 2 and 3 compatibility utilities
starlette 0.13.2 The little ASGI library that shines.
tenacity 6.2.0 Retry code until it succeeds
└── six >=1.9.0
uvicorn 0.11.5 The lightning-fast ASGI server.
├── click >=7.0.0,<8.0.0
├── h11 >=0.8,<0.10
├── httptools >=0.1.0,<0.2.0
├── uvloop >=0.14.0
└── websockets >=8.0.0,<9.0.0
```
poetry does not seem to think uvicorn is a dependency of fastapi or starlette

tomjw64: It might be a peer dependency rather than a child?

backend/backend.dockerfile
```diff
@@ -1,29 +1,61 @@
...
+ && apt-get autoremove -y
```
tomjw64, line 8: Also both ampersand lines could use one additional space of indentation if formatting is to be consistent.

backend/app/pyproject.toml
m1yag1: ah that makes sense. i don't know why i thought it was a part those. as i rack my brain i remember it's added to the project b/c uvloop is a much faster event loop implementation for asyncio.

backend/backend.dockerfile
m1yag1: I actually had it there first and then moved it up b/c i thought it would be good to just install it as part of base and not worry about it. However, you comment has me thinking that multi-stage builds are like JIT so it should go back. I'll make that change.

TylerZeroMaster (approved):

PR #117, opened 2020-06-15
Complete implementation of check-feed task

rnathuji:


rnathuji:
There are a couple aspects of this PR I want to point out / discuss:
1. Timer periodicity
2. Behavior when there is no work to do

### Timer periodicity
The goal is to use a periodic timer to trigger the distribution pipeline. I've currently set this to be hourly. A tradeoff here is if we set it too low, there's an increased likelihood of a book being built in concurrent pipelines due to the use of the `.complete` file written at the end of `upload-book` to decide whether a book has already been built (our replicator protection should prevent this from being a problem in terms of overwriting files, but in general the reliance on a `.complete` file subject to eventual consistency means the problem of potential concurrent builds isn't completely avoidable in any case). If it's too high, it will take longer for a new code version to process an entire feed. 

### Behavior when there is no work to do
The `check-feed` task will return a non-zero exit code causing the pipeline to not invoke subsequent tasks when it finds all books have been built. A side effect is that builds which have no work will show as failed in Concourse. I think this is mostly cosmetic, but the red status may be annoying as a false negative. If we want to avoid it, I believe the only way is to separate a job that checks the feed from one that processes entries via a resource which triggers the latter.

rnathuji:
Through various discussions, we identified a modified approach for the pipeline which lets us take advantage of parallelism (so we can complete a build of all books within 24h) while also avoiding the need for strong coordination. Specifically, we will structure the pipeline as two jobs:

1. `feeder` job: Decides what should be built based upon the entries in the accepted books JSON list and current state in S3
2. `bakery` job: Existing pipeline which can build a single book

The feeder job will be triggered by a time resource configured with a desired tick interval in the pipeline. It will submit multiple book entries (up to some parameterized limit) by writing to a versioned S3 file. Each write will be a JSON object with the required book parameters.

The bakery job will be triggered by the S3 resource. When there are multiple versions of the file available, Concourse will run concurrent builds of this job.

bakery/src/pipelines/distribution.js
```diff
@@ -25,24 +28,49 @@ const pipeline = (env) => {
...
+      source: {
+        interval: env.PIPELINE_TICK_INTERVAL
+      }
     }
   ]
 
+  const feederJob = {
+    name: 'feeder',
+    plan: [
+      { get: 'ticker', trigger: true },
```
tomjw64, line 53: We will miss books in the queue if we don't use `version: every` here
```suggestion
      { get: 'ticker', trigger: true, version: 'every' },
``

File: bakery/src/pipelines/distribution.js [hunk 2]
```diff
@@ -25,24 +28,49 @@ const pipeline = (env) => {
...
+      taskCheckFeed({
+        awsAccessKeyId: awsAccessKeyId,
+        awsSecretAccessKey: awsSecretAccessKey,
+        feedFileUrl: env.FEED_FILE_URL,
+        versionedBucketName: env.S3_VERSIONED_BUCKET,
+        versionedFile: versionedFile,
+        codeVersion: codeVersionFromTag,
+        maxBooksPerRun: env.MAX_BOOKS_PER_TICK,
+        image: { tag: env.IMAGE_TAG }
```
tomjw64, line 62:
```suggestion
        image: { tag: lockedTag }
```

bakery/env/local.json
```diff
@@ -5,5 +5,9 @@
   "S3_ACCESS_KEY_ID": "",
   "S3_SECRET_ACCESS_KEY": "",
   "S3_DIST_BUCKET": "",
-  "VERSIONED_FILE": ""
+  "S3_VERSIONED_BUCKET": "",
+  "VERSIONED_FILE": "",
+  "FEED_FILE_URL": "",
+  "PIPELINE_TICK_INTERVAL": "",
+  "MAX_BOOKS_PER_TICK": ""
```
tomjw64, line 12: We decided that if we needed more responsiveness, we should add a `pending` state so that we could queue books at some shorter time interval, but that if we didn't the equivalent was queuing all books at a large time interval. If this is the case, should we not need `MAX_BOOKS_PER_TICK` (or we should set it to a value greater than the number of books we expect on the list) until such time arrives that we need a `pending` state and higher responsiveness?

bakery/env/local.json
```diff
@@ -5,5 +5,9 @@
   "S3_ACCESS_KEY_ID": "",
   "S3_SECRET_ACCESS_KEY": "",
   "S3_DIST_BUCKET": "",
-  "VERSIONED_FILE": ""
+  "S3_VERSIONED_BUCKET": "",
```
tomjw64, line 8: Rename to `QUEUE_STATE_BUCKET`? :stuck_out_tongue:

bakery/env/local.json
```diff
@@ -5,5 +5,9 @@
...
+  "VERSIONED_FILE": "",
```
tomjw64, line 9: Rename to `QUEUE_FILENAME` or similar? :stuck_out_tongue:

bakery/src/pipelines/distribution.js
rnathuji: This is for the timer tick versus the file, though. I think the only reason I didn't include the "every" here was in case it was delayed and there were two "ticks" too close (maybe this isn't a real issue with the time resource, though?)

The file resource version does have "every" on line 70 below.

bakery/env/local.json
rnathuji: I just piped it through for now since if nothing else it is useful for development / testing already to be able to set this value. I agree that initially the production value for this setting should be something high, but left it blank for now as I wasn't sure of a good value just yet. I can set something in `prod.json`, though, if there's a number you think makes sense to put for now.

bakery/env/local.json
rnathuji: Yeah, let's make this line up with the CF code so it's clear. I'll use your suggested values for this and `VERSIONED_FILE`.

bakery/src/pipelines/distribution.js
tomjw64: Oops! Wrong resource! There is `version: every` on the resource that actually matters!

bakery/env/local.json
tomjw64: As long as we know to set it high for prod, I'm :ok_hand:

PR #120, opened 2020-06-26
Prep for npm packaging

tomjw64:


bakery/.npmignore
```diff
@@ -0,0 +1,6 @@
+src/tests
+src/scripts
+**/__pycache__/
+credentials.yml.template
```
rnathuji, line 4: Similar to the `distribution-feed.json`, can we delete `credentials.yml.template` altogether as well (I think it's a remnant from when we migrated code from `concourse-pipelines` and don't believe it's used anymore)?

PR #127, opened 2020-07-01
Add XHTML validation to bakery pipelines

rnathuji:


rnathuji:
Note / Feedback request: The PR currently only enables 2/3 checks available from our `xhtml-validator` (see [comment](https://github.com/openstax/cnx/issues/1010#issuecomment-652556460) in parent issue). I'm not sure if the observed failures on `duplicate-id` checks are something we'd expect today, or if they should be addressed in any book(s) that have them and this PR needs to enforce that validation check as well.

rnathuji:
@tomjw64 - Just to clarify your comment, are you saying perhaps we should just call validate once per pipeline -- say after jsonify-book in distribution pipeline and mathify in pdf pipeline?

tomjw64:
That's correct

rnathuji:
So the current version of this went for the "put it everywhere" end of the spectrum based upon the solution approach outlined in the task description. I don't have a strong reason for not only checking once (near the end) of each pipeline other than the benefit of failing as early as possible. I guess in general it is theoretically possible that things could pass after disassemble but be broken earlier -- but I don't know 1) If it's realistically ever possible; 2) Even if it were whether we'd care since the issue wouldn't be observable in any generated artifacts.

Is the main concern / tradeoff of minimizing invocations just reducing the build time overheads?

tomjw64:
My main concern is indeed the overhead multiplied by however many books we build. And ideally, since this should be a rare error case, to me it makes sense to avoid starting up 3 extra processes during a book build and simply error and investigate when we do not expect it to happen.
I am thinking along the same lines as the points you outlined. If the content we are shipping off downstream validates (but maybe not necessarily some intermediate state), I would consider that a success.

PR #132, opened 2020-07-07
adding verification for second error message in colid field - incorrect colid

omehes:
verification for incorrect colid error message


omehes:
Note, that I had to add sleep time to ```def click_create_button(self):``` because the test was failing and passing due to being to fast.

therealmarv:
It seems `ci/circleci: integration-test` is still failing (click on Details).

omehes:
that is weird. that file (test_incorrect_col_id_error) does not exist anymore. I deleted it before commiting and pushing the branch. the file where the new test is, is "test_modal_field_errors.py" and that passed.

omehes:
@therealmarv I did some little changes and had to force push the changes but now it looks good. Integration tests passed...

PR #134, opened 2020-07-09
update canonical-book-list.json

ckline-tryptic:


rnathuji:
Should have left a note yesterday on this PR: @ckline-tryptic and I discussed the changes a bit on 7/9, and the modified structure may not be necessary after all with the direction he's going to to try and take for [cnx 1039](https://github.com/openstax/cnx/issues/1039). I suggest we place this PR on the back burner or in draft state for the time being. We can later close or merge it depending on what he finds during implementation of the related issue.

PR #140, opened 2020-07-17
cops e2e test - adding classes

omehes:
adding classes for 4 input fields
adding class for status message to index.vue


omehes:
@therealmarv I cannot find better locator for the job status field. Spending too much time on this. xpath locator works. If you have time to review this and comment.

therealmarv (approved):
I think if this works we should leave it this way. I know the xpath locator looks more complicated but maybe we can improve it on another PR in the future.

PR #146, opened 2020-07-24
update physics to college-physics in dropdown

brittany-johnson:
We updated the book name `physics` to `college-physics` in `book.txt`


therealmarv:
@brittany-johnson do you need that change to be deployed on production COPS? When yes, when? ASAP?

PR #147, opened 2020-07-24
Re-enable duplicate id validation

tomjw64:
Switches out the links-to-duplicate-id for the duplicate-id validation (if there are no duplicate ids, there can be no links to duplicate ids!) My hunch is that with the new version of cnx-epub, this should pass in most cases now. (But I could be wrong, in which case, we should revert it again)

### DIFF
--- a/bakery/src/pipelines/distribution.js
+++ b/bakery/src/pipelines/distribution.js
@@ -86,7 +86,8 @@ const pipeline = (env) => {
       taskValidateXhtml({
         image: { tag: lockedTag },
         inputSource: 'jsonified-book',
-        inputPath: 'jsonified/*@*.xhtml'
+        inputPath: 'jsonified/*@*.xhtml',
+        validationName: 'duplicate-id'
       }),
       taskUploadBook({
         distBucket: env.S3_DIST_BUCKET,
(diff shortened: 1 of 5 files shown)


brittweinstein:
@tomjw64 please fill out the milestone, estimate and release

bakery/src/pipelines/distribution.js
```diff
@@ -86,7 +86,8 @@ const pipeline = (env) => {
       taskValidateXhtml({
         image: { tag: lockedTag },
         inputSource: 'jsonified-book',
-        inputPath: 'jsonified/*@*.xhtml'
+        inputPath: 'jsonified/*@*.xhtml',
+        validationName: 'duplicate-id'
```
rnathuji, line 90: This isn't directly related to this change, but it occurred to me when we eventually turn on all the checks the `validationName` would be an empty string with the current implementation of `xhtml-validator`. Should we create an issue to make a change there so it's `all` or something more explicit for later?

tomjw64: Well, some validations are subsets of others. If duplicate-id passes, link-to-duplicate-id must also pass. I think it would be handy to take an argument list of the validations to run. (But I agree with 'all', if we don't really care about doing a little bit of extra work twice, which I don't really, but a list may more flexible)

PR #148, opened 2020-07-28
Remove credentials from local.json env settings

rnathuji:
In order to avoid the risk of committing secrets, when building a
pipeline for local environments we'll grab the AWS credentials using
the corresponding environment variables typically used for AWS CLI.

### DIFF
--- a/bakery/src/tests/test_cli.js
+++ b/bakery/src/tests/test_cli.js
@@ -66,13 +66,42 @@ test('build pipelines', async t => {
           'pipeline',
           pipeline,
           env
-        ]))
+        ],
+        {
+          // Include credentials in environment for local pipelines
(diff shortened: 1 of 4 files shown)


bakery/src/tests/test_cli.js
```diff
@@ -66,7 +66,17 @@ test('build pipelines', async t => {
...
+        {
+          env: {
+            ...process.env,
+            ...{
+              AWS_ACCESS_KEY_ID: 'accesskey',
+              AWS_SECRET_ACCESS_KEY: 'secret'
+            }
+          }
+        }
+        ))
```
tomjw64, line 79: Should we add a test that asserts an error occurs if the vars are not provided?

PR #149, opened 2020-07-28
add external book uuid linking

ckline-tryptic:
Leaves unit testing in test_bakery_script.py until we convert to a package.

### DIFF
--- a/bakery/src/cli/execute.js
+++ b/bakery/src/cli/execute.js
@@ -325,6 +325,53 @@ const tasks = {
       }
     }
   },
+  'link-extras': (parentCommand) => {
+    const commandUsage = 'link-extras <collid> <server>'
+    const handler = async argv => {
+      const buildExec = path.resolve(BAKERY_PATH, 'build')
+
+      const imageDetails = imageDetailsFromArgs(argv)
+      const taskArgs = imageDetails == null
+        ? []
+        : [`--taskargs=${JSON.stringify({
+          ...imageDetails, ...{ server: argv.server }
+        })}`]
+
+      const taskContent = execFileSync(buildExec, ['task', 'link-extras', ...taskArgs])
+      const tmpTaskFile = tmp.fileSync()
+      fs.writeFileSync(tmpTaskFile.name, taskContent)
+
+      const tmpBookDir = tmp.dirSync()
+      fs.writeFileSync(path.resolve(tmpBookDir.name, 'collection_id'), argv.collid)
+
+      const dataDir = path.resolve(argv.data, argv.collid)
+
+      await flyExecute([
+        '-c', tmpTaskFile.name,
+        `--input=book=${tmpBookDir.name}`,
+        input(dataDir, 'assembled-book'),
+        output(dataDir, 'linked-extras')
+      ], { image: argv.image, persist: argv.persist })
+    }
+    return {
+      command: commandUsage,
+      aliases: 'l',
+      describe: 'amend external book links',
+      builder: yargs => {
+        yargs.usage(`Usage: ${process.env.CALLER || `$0 ${parentCommand}`} ${commandUsage}`)
+        yargs.positional('collid', {
+          describe: 'collection id of collection to work on',
+          type: 'string'
+        }).positional('server', {
+          describe: 'archive server',
+          type: 'string'
+        })
+      },
+      handler: argv => {
+        handler(argv).catch((err) => { console.error(err); process.exit(1) })
+      }
+    }
+  },
   bake: (parentCommand) => {
     const commandUsage = 'bake <collid> <recipefile> <stylefile>'
     const handler = async argv => {
@@ -354,7 +401,7 @@ const tasks = {
       await flyExecute([
         '-c', tmpTaskFile.name,
         `--input=book=${tmpBookDir.name}`,
-        input(dataDir, 'assembled-book'),
+        input(dataDir, 'linked-extras'),
         `--input=cnx-recipes-output=${tmpRecipesDir.name}`,
         output(dataDir, 'baked-book')
       ], { image: argv.image, persist: argv.persist })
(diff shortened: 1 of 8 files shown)


bakery/src/cli/execute.js
```diff
@@ -325,6 +325,47 @@ const tasks = {
       }
     }
   },
+  link: (parentCommand) => {
+    const commandUsage = 'link <collid> <canonicallist>'
```
rnathuji, line 329: Isn't the only argument to this command `collid`?

bakery/src/cli/execute.js
```diff
@@ -325,6 +325,47 @@ const tasks = {
       }
     }
   },
+  link: (parentCommand) => {
```
rnathuji, line 328: To be consistent with the other commands such as `bake-meta` etc., I think (assuming we leave this called `link-book`), we should change this to `'link-book'`

bakery/src/cli/execute.js
```diff
@@ -325,6 +325,47 @@ const tasks = {
...
+    const commandUsage = 'link <collid> <canonicallist>'
+    const handler = async argv => {
+      const buildExec = path.resolve(BAKERY_PATH, 'build')
+
+      const imageDetails = imageDetailsFromArgs(argv)
+      const taskArgs = imageDetails == null
+        ? []
+        : [`--taskargs=${JSON.stringify(imageDetails)}`]
+      const taskContent = execFileSync(buildExec, ['task', 'link-book', ...taskArgs])
```
rnathuji, line 337: The name `link-book` seems really vague to me. I know we brainstormed some potential options in Slack, and perhaps it's worth revisiting that conversation? (names are hard and subjective, I know, and perhaps `link-book` is the best option we can come up with)

bakery/src/scripts/link-book.py
```diff
@@ -0,0 +1,61 @@
...
+    for node in doc.xpath(
+        '//x:a[@href and starts-with(@href, "/contents/m")]',
+        namespaces={"x": "http://www.w3.org/1999/xhtml"},
+    ):
+        # find legacy module id
+        link = node.attrib["href"]
+        module_id = link.lstrip("/contents/")
+
+        # get target module uuid
+        req = requests.get(f"https://archive.cnx.org/content/{module_id}")
```
rnathuji, line 35: Should we hard code this URL for the archive endpoint within the script, or parameterize that as a command arg or environment variable so we can adjust the endpoint at the task / pipeline level if needed?

bakery/src/scripts/link-book.py
```diff
@@ -0,0 +1,61 @@
...
+        if req.status_code != 200:
+            continue
```
rnathuji, line 37: Instead of masking potential errors, should we just `req.raise_for_status()` and let it error out? (the idea being it's better to fail than lead to non-deterministic builds) Or am I misunderstanding something?

bakery/src/scripts/link-book.py
```diff
@@ -0,0 +1,61 @@
...
+        module_uuid = req.url.split("/")[-1]
+
+        # get list of books containing module
+        req = requests.get(f"https://archive.cnx.org/extras/{module_uuid}")
```
rnathuji, line 41: Should we check the status_code / do a `req.raise_for_status()` on this call too?

bakery/src/scripts/link-book.py
```diff
@@ -0,0 +1,61 @@
...
+        content = json.loads(req.content)
```
rnathuji, line 42: Minor note: An alternative here would be to use `req.json()`

bakery/src/scripts/link-book.py
```diff
@@ -0,0 +1,61 @@
...
+            book["ident_hash"].split("@")[0]: book["title"] for book in content["books"]
+        }
+
+        # match uuid in canonical book list
+        for book_uuid in canonical_ids:
+            if book_uuid not in containing_books:
+                continue
+            node.attrib["href"] = f"/contents/{module_uuid}"
+            node.attrib["data-book-uuid"] = book_uuid
+            break
```
rnathuji, line 53: I can't find it anywhere now in the swarm doc notes, but I believe there was an error case we wanted to watch for in this script (see related notes in [cnx 1026](https://github.com/openstax/cnx/issues/1026) though I don't think they reflect the most recent convos on this point). 

IIRC, I think we ended up with wanting to error when there are multiple books returned from extras AND none of them are in our canonical list. The idea being if there is only one book returned from extras and it's not in our list it won't matter, but otherwise if we don't get a hit we should fail loudly. Does that ring a bell / sound right to you? Currently it looks like we won't modify the link if there's no hit in the list (?)

bakery/src/tests/test_cli.js
```diff
@@ -174,6 +174,17 @@ test('stable flow in pdf and distribution pipeline', async t => {
   const assembleMetaResult = await completion(assembleMeta)
   t.truthy(fs.existsSync(`${outputDir}/${bookId}/assembled-book-metadata/${bookId}/collection.assembled-metadata.json`), formatSubprocessOutput(assembleMetaResult))
 
+  const link = spawn('node', [
+    'src/cli/execute.js',
+    ...commonArgs,
+    '--image=localhost:5000/openstax/cops-bakery-scripts:test',
+    'link',
+    bookId,
+    'src/scripts/canonical-book-list.json'
```
rnathuji, line 183: I'm confused -- why are we trying to passing this when the task is grabbing from the files in the container via `/code/scripts/canonical-book-list.json`?  🤷‍♂️

bakery/src/tests/test_cli.js
```diff
@@ -174,6 +174,17 @@ test('stable flow in pdf and distribution pipeline', async t => {
   const assembleMetaResult = await completion(assembleMeta)
   t.truthy(fs.existsSync(`${outputDir}/${bookId}/assembled-book-metadata/${bookId}/collection.assembled-metadata.json`), formatSubprocessOutput(assembleMetaResult))
 
+  const link = spawn('node', [
```
rnathuji, line 177: General question as far as adding this test case: Do you know whether the test book results in queries to archive? (e.g. are there external book links in `col30149`) If so, we may want to think about whether to include as part of these.

PR #152, opened 2020-07-31
Add linting for bakery Python scripts

rnathuji:


tomjw64:
Looks like a job for [autopep8](https://pypi.org/project/autopep8/)!

rnathuji:
@tomjw64 - Thanks for the pointer to autopep8. Unfortunately, even with a couple of `-a`'s thrown in it had issues with some of our long lines. But things are linted now in any case.

PR #156, opened 2020-08-04
Add placeholder pipeline for gdoc creation

tomjw64:


bakery/src/pipelines/gdoc.js
```diff
@@ -0,0 +1,104 @@
...
+    plan: [
+      { get: 's3-queue', trigger: true, version: 'every' },
+      { get: 'cnx-recipes-output' },
+      taskDequeueBook({
+        queueFilename: queueFilename,
+        image: { tag: lockedTag }
+      }),
+      taskFetchBook({ image: { tag: lockedTag } }),
+      taskAssembleBook({ image: { tag: lockedTag } }),
+      taskAssembleBookMeta({ image: { tag: lockedTag } }),
```
rnathuji, line 79: Just to check: It's probably not worth incorporating the recently added `taskLinkExtras` here, since I believe the results of that link fixing are still going to be broken in this context, and the gdoc specific link patching will be a separate task altogether per [cnx 1115](https://github.com/openstax/cnx/issues/1115), correct?

bakery/src/pipelines/gdoc.js
```diff
@@ -0,0 +1,104 @@
...
+  const taskBakeBookMeta = require('../tasks/bake-book-metadata')
+  const taskChecksumBook = require('../tasks/checksum-book')
+  const taskDisassembleBook = require('../tasks/disassemble-book')
+  const taskJsonifyBook = require('../tasks/jsonify-book')
+  const taskValidateXhtml = require('../tasks/validate-xhtml')
+
+  const awsAccessKeyId = env.ENV_NAME === 'local' ? env.S3_ACCESS_KEY_ID : '((aws-sandbox-secret-key-id))'
+  const awsSecretAccessKey = env.ENV_NAME === 'local' ? env.S3_SECRET_ACCESS_KEY : '((aws-sandbox-secret-access-key))'
+  const codeVersionFromTag = env.IMAGE_TAG || 'version-unknown'
+  const queueFilename = `${codeVersionFromTag}.${env.QUEUE_FILENAME}`
```
rnathuji, line 17: General question related to the combination of `S3_QUEUE_STATE_BUCKET` and `QUEUE_FILENAME` in this pipeline. Do we want the queue files for gdoc to live in a separate bucket from web-hosting builds, or the same bucket but different name structure for the file, or some combination of both, or ...?

bakery/src/pipelines/gdoc.js
tomjw64: That seems correct to me. I'll add another comment add a placeholder for that as well.

bakery/src/pipelines/gdoc.js
tomjw64: Seems like there's no specific decision, so I left it as is, since we determined the functionality will at least be similar. I am aiming to kick the can down the road on this one.

PR #158, opened 2020-08-05
refactor link_extras, fix requests.json

ckline-tryptic:


bakery/src/scripts/link_extras.py
```diff
@@ -10,62 +10,102 @@
...
+
+    match = None
+    for book_uuid in canonical_ids:
+        if book_uuid in containing_books:
+            match = book_uuid
+            break
+
+    # handle multiple containing books with no canonical match
+    if match == None:
+        raise Exception("Multiple containing books, no canonical match!")
```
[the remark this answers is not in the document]
tomjw64: Adding
```
if len(containing_books) == 0:
    raise Exception('No containing books')
```
to the top of the function may help.

bakery/src/scripts/link_extras.py
```diff
@@ -10,62 +10,102 @@
...
+        book["ident_hash"].split("@")[0]: book["title"]
+        for book in content["books"]
+    }
+
+
+def match_canonical_book(canonical_ids, containing_books):
+    """match uuid in canonical book list"""
+    if len(containing_books) == 1:
+        [match] = containing_books
+        return match
```
tomjw64, line 58: Kinda unimportant, but binding a new variable doesn't really help us or reduce any complexity for us here. Destructuring the dict into a list masks the fact that containing_books is a dict.
`return list(containing_books.keys())[0]` could be used. Or clearly `return containing_books[0]` could be used if this was simply a list (depending on how you respond to my above comment.

bakery/src/scripts/link_extras.py
```diff
@@ -10,62 +10,102 @@
...
+
+    match = None
+    for book_uuid in canonical_ids:
+        if book_uuid in containing_books:
+            match = book_uuid
+            break
```
tomjw64, line 64:
```suggestion
match = next(filter(lambda uuid: uuid in containing_books, canonical_ids))
```
is an equivalent expression if you think functions read better, but totally up to you.

bakery/src/scripts/link_extras.py
```diff
@@ -10,62 +10,102 @@
...
+
+
+def get_containing_books(server, module_uuid):
+    """get list of books containing module"""
+    req = requests.get(f"https://{server}/extras/{module_uuid}")
+    req.raise_for_status()
+
+    content = req.json()
+
+    return {
```
tomjw64, line 48: We never really use the values of this dict, only the keys. Do we plan to use the values? Is it worth making this a list?

bakery/src/scripts/link_extras.py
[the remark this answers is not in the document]
tomjw64: Although, this may also be a corner case we don't care about, since no module should be standalone without a parent collection.

PR #159, opened 2020-08-10
Add docx conversion task

tomjw64:


therealmarv:
Things probably missing:
- Inter-module linking

I'm not sure if resources are handled the right way (did you test with images?)

Reference LOC in the bash script regarding inter module links and resources https://github.com/openstax/gdoc-spike/blob/5e9fb4d8d02d5cc0260c5801f571704f17e675a3/script/do-calculus.bash#L18-L45

FYI: the table wrapping and custom reference docx will change. Maybe also a new transformation step will get inside but this relies on final decision by Andy and Bey sometime this week.

tomjw64:
@therealmarv Would you be comfortable making that change in the bakery task code, or would you like to wait to merge this one in?

therealmarv:
My work will be in done in gdocs-spike repo but once I've got a final decision and okay from Bey and Andy I can modify the bakery task code with the improvements and a new PR  based on your work 🙂  (so my changes can be for now independent from the general task execution code here).

PR #163, opened 2020-08-12
Catch empty generator exception

tomjw64:
Noticed a job failed QA pipeline with a StopIteration exception. This will catch it a re-raise a more descriptive error. Still a valid failure mode, however.


ckline-tryptic:
Maybe we should change the message to give some insight into why it is a problem. Maybe something like: "Ambiguous external module. No canonical match."?

PR #165, opened 2020-08-13
Check mem limit

philschatz:
and warn if it is too small.

This is an add-on to #164 in case macOS developers miss the documentation

### DIFF
--- a/bakery/src/cli/execute.js
+++ b/bakery/src/cli/execute.js
@@ -12,7 +12,7 @@ tmp.setGracefulCleanup()
 const sleep = ms => new Promise(resolve => setTimeout(resolve, ms))
 
 const checkDockerMemoryLimit = () => {
-  const dockerSettingsPath = path.join(process.env['HOME'], 'Library/Group Containers/group.com.docker/settings.json')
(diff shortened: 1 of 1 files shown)


bakery/src/cli/execute.js
```diff
@@ -11,6 +11,23 @@ tmp.setGracefulCleanup()
 
 const sleep = ms => new Promise(resolve => setTimeout(resolve, ms))
 
+const checkDockerMemoryLimit = () => {
+  const dockerSettingsPath = path.join(process.env.HOME, 'Library/Group Containers/group.com.docker/settings.json')
+  if (fs.existsSync(dockerSettingsPath)) {
+    const settings = JSON.parse(fs.readFileSync(dockerSettingsPath))
+    if (!(settings.memoryMiB > 4096)) {
+      console.warn(`
```
tomjw64, line 19: you can use dedent so you don't break indendation levels with the template string

philschatz:
Yeah, the app only exists for OSX and Windows users and I do not know how to figure out where the Windows version of the app stores this information. 

To find out where the settings were, I used the **File Activity** mode in Instruments.app (inside XCode’s directory) and recorded myself opening up Docker’s Preferences page.

image

PR #168, opened 2020-08-18
add tests for link-extras step

ckline-tryptic:


bakery/src/scripts/link_extras.py
```diff
@@ -1,6 +1,6 @@
 """
 Replaces legacy module ids in links to external modules with
-uuids from the target module and corresponding canonical book.
+uuids from the target module and corresponding canonical book       .
```
tomjw64, line 3:
```suggestion
uuids from the target module and corresponding canonical book.
```

tomjw64:
Only the main method is not covered by tests. Mostly worthless to get coverage on that. Approved.

PR #169, opened 2020-08-18
Prod + Staging envs yield respective secrets

tomjw64:


bakery/src/tests/test_cli.js
```diff
@@ -69,13 +90,13 @@ test('build pipelines', async t => {
         ],
         {
           // Include credentials in environment for local pipelines
```
rnathuji, line 92: Minor: I think this comment can be removed

bakery/env/prod.json
```diff
@@ -2,10 +2,12 @@
...
+  "S3_DIST_BUCKET": "web-hosting-content-gatekeeper",
+  "S3_QUEUE_STATE_BUCKET": "web-hosting-content-queue-state",
+  "QUEUE_FILENAME": "distribution-queue.json",
   "FEED_FILE_URL": "https://raw.githubusercontent.com/openstax/content-manager-approved-books/master/approved-books.json",
-  "PIPELINE_TICK_INTERVAL": "",
-  "MAX_BOOKS_PER_TICK": ""
+  "PIPELINE_TICK_INTERVAL": "24h",
+  "MAX_BOOKS_PER_TICK": "50",
+  "DIST_BUCKET_AKI_SECRET_NAME": "((prod-web-hosting-content-gatekeeper-access-key-id))",
+  "DIST_BUCKET_SAK_SECRET_NAME": "((prod-web-hosting-content-gatekeeper-secret-access-key))"
```
rnathuji, line 12: Just to check my understanding: The idea is the same credentials get used for all three pipelines with this change, correct (though perhaps meant to be temporary since the name is specific to `dist`)? If so, will we need to extend the corresponding IAM user policy so that the production PDF pipeline (which will I believe now incorporate `((prod-web-hosting-content-gatekeeper-access-key-id))`) can write to its current s3 bucket (which I think is currently actually in our sandbox AWS, but I expect will move at some point)?

bakery/src/tests/test_cli.js
tomjw64: I'm not sure I understand. The reason those creds need to be there is still for the local pipeline to not fail, right?

bakery/env/prod.json
tomjw64: That was not the idea. I think there's some changes to be made for how these env files are handled in the future, because right now, the variables used by pdf vs distribution/gdoc are exclusive to those pipelines and I don't believe there is any intermixing. Because of this, this secret name will not and should not be embedded within any PDF pipeline, only gdoc/dist, since they share the same target bucket.

Additionally staging/prod names are specific to pdf COPS, but we've repurposed them to mean sandbox/prod for the purposes of distribution/gdoc pipeline deployments, which to me is gonna end up being really confusing. Hopefully this is good enough for today, since it seems to be secure enough and we are already behind the deadline for moving content-distribution to production infrastructure. I'm open to ideas on how to restructure the env distinctions going forward. Maybe we should make a investigation/decomp card?

I think what you've identified is some additional infrastructure we will need to have the PDF pipelines also target production account s3 buckets (which will make use of a different IAM User and different credentials, I'm guessing, since the pdf pipeline will be accessing different resources than the other pipelines)

bakery/src/tests/test_cli.js
rnathuji: Oh, perhaps I didn't read this right when looking at it earlier. Is this now passing those for all environments (e.g. the change to remove `env === 'local'`)?

tomjw64: correct, it should be fine to include the environment variables for any pipeline, although, I'm just now realizing I should add a test to check that non-local pipelines do not use the env vars if present (since they would be inlined and readable with fly)

bakery/env/prod.json
rnathuji: Agreed on the general statements above as far as deferring thinking about options to restructure / cleanup in a separate card. Mostly for this one, I'm wondering whether once merged if it would break PDF production since `./build pipeline pdf prod` with the changes produces a modified s3 resource definition for me 🤷‍♂️ :

```
  - name: s3
    type: s3
    source:
      bucket: ce-pdf-spike
      access_key_id: ((prod-web-hosting-content-gatekeeper-access-key-id))
      secret_access_key: ((prod-web-hosting-content-gatekeeper-secret-access-key))
      skip_download: true
```

tomjw64: yeah, that would break stuff. that should not be present in the pdf pipeline. I will correct that.

tomjw64: good catch

bakery/src/tests/test_cli.js
tomjw64: Okay, added that test.

bakery/env/prod.json
tomjw64: Added in keys specifically for pdf s3 target secret names (just used the sandbox keys for both for now, since both staging + prod COPS target a sandbox bucket - this might be the temporary piece you are referring to :smile: )

PR #171, opened 2020-08-19
Add upload-docx task

rnathuji:


bakery/src/tasks/upload-docx.js
```diff
@@ -0,0 +1,53 @@
...
+          echo "$GOOGLE_SERVICE_ACCOUNT_CREDENTIALS" > /tmp/service_account_credentials.json
+          collection_id="$(cat book/collection_id)"
+          book_legacy_version="$(cat book/version)"
+          docx_dir="docx-book/$collection_id/docx"
+          book_metadata="docx-book/$collection_id/raw/metadata.json"
+          book_title="$(cat $book_metadata | jq -r '.title')"
+          upload-docx "$docx_dir" "$book_title" "${parentGoogleFolderId}" /tmp/service_account_credentials.json
+          complete_filename=".$collection_id@$book_legacy_version.complete"
+          date -Iseconds > "/tmp/$complete_filename"
+          aws s3 cp "/tmp/$complete_filename" "s3://${queueStateBucket}/${codeVersion}/$complete_filename"
```
tomjw64, line 45: Do gdoc pipeline and web hosting pipeline share the same queue state file names per version? If we spin up each pipeline with the same version, what prevents them from affecting each other's visible states? I'm just now realizing the check-feed job that I pushed up may share this issue :fearful:

bakery/src/tasks/upload-docx.js
```diff
@@ -0,0 +1,53 @@
...
+        AWS_SECRET_ACCESS_KEY: `${awsSecretAccessKey}`
+      },
+      inputs: [
+        { name: 'book' },
+        { name: 'docx-book' }
+      ],
+      run: {
+        path: '/bin/bash',
+        args: [
+          '-ce', // FIXME: Not including -x to avoid displaying credentials when writing to file.
```
tomjw64, line 34: could you `set -x` after the credentials line?

bakery/src/scripts/upload_docx.py
```diff
@@ -0,0 +1,142 @@
...
+          f"name='{folder_name}'",
+        fields='files(id)'
+    ).execute()
+
+    items = results.get("files", [])
+
+    if len(items) > 1:
+        raise Exception(f"Multiple folders with name '{folder_name}' found.")
+
+    if items:
```
tomjw64, line 26: Had to double check that `[]` is falsey. Nice :smile:

bakery/src/tasks/upload-docx.js
rnathuji: Good catch (I can update the `check-feed` side in tandem with this). We should probably tease apart the `QUEUE_FILENAME` and `FEED_FILE_URL` values for `gdoc` too, no? (so they use the same `S3_QUEUE_STATE_BUCKET`, but we can add `GDOC_QUEUE_FILENAME` and `GDOD_FEED_FILE_URL`).

bakery/src/tasks/upload-docx.js
rnathuji: Oh, man -- of course. Thanks!

bakery/src/tasks/upload-docx.js
rnathuji: I added a prefix to the state files so we can keep things separate. Just to be sure, I tested by running concurrent pipelines of dist and gdoc that used the same tag, and they didn't trample each other.

PR #176, opened 2020-08-26
Bn/cnx#1123/test coverage s3 upload

c0ldcalamari:


bakery/src/tests/test_bakery_scripts.py
```diff
@@ -857,3 +859,245 @@ def test_gdocify_book(tmp_path, mocker):
...
+
+    book_dir = tmp_path / "col11762"
+    book_dir.mkdir()
+
+    resources_dir = book_dir / "resources"
+    resources_dir.mkdir()
+    resource = resources_dir / "fffe62254ef635871589a848b65db441318171eb.json"
+
+    test_data_one = open(test_resource, "rb").read()
+    resource.write_bytes(test_data_one)
```
rnathuji, line 878: Perhaps I'm confused, but is this going to write the data from `src/test/data/fffe62254ef635871589a848b65db441318171eb` to `{tmp_path}/col11762/resources/fffe62254ef635871589a848b65db441318171eb.json`?

bakery/src/tests/test_bakery_scripts.py
```diff
@@ -857,3 +859,245 @@ def test_gdocify_book(tmp_path, mocker):
...
+            'ContentType': 'image/jpeg',
+            'Key': key,
+        }
+    )
+
+    s3_stubber.activate()
+
+    from unittest.mock import MagicMock
+    mocked_session = boto3.session.Session
+    mocked_session.client = MagicMock(return_value=s3_client)
```
rnathuji, line 923: Minor: I believe you can avoid the inline import above by using `mocker.MagicMock` via the pytest fixture here.

bakery/src/tests/test_bakery_scripts.py
```diff
@@ -857,3 +859,245 @@ def test_gdocify_book(tmp_path, mocker):
...
+
+    mocker.patch(
+        'boto3.session.Session',
+        mocked_session
+    )
+
+    mocker.patch(
+        'sys.argv',
+        ['',
+         TEST_DATA_DIR,
```
rnathuji, line 933: Did you mean to pass `tmp_path` here instead? (there's some input data written out to it above, but then you're passing the data path directly here 🤷‍♂️ )

bakery/src/tests/test_bakery_scripts.py
```diff
@@ -857,3 +859,245 @@ def test_gdocify_book(tmp_path, mocker):
...
+    )
+
+    s3_stubber.add_response(
+        "put_object",
+        {},
+        expected_params={
+            'Body': botocore.stub.ANY,
+            'Bucket': dist_bucket,
+            'ContentType': 'application/json',
+            'Key': 'master/resources/fffe62254ef635871589a848b65db441318171eb-unused.json',
```
rnathuji, line 904: Minor: Looks like the CI linter is complaining about this line being too long

bakery/src/tests/test_bakery_scripts.py
```diff
@@ -857,3 +859,245 @@ def test_gdocify_book(tmp_path, mocker):
...
+    aws_key = 'dummy-key'
+    aws_secret = 'dummy-secret'
+
+    upload_resource = copy_resources_s3.check_s3_existence(
+        aws_key, aws_secret,
+        bucket, resource_for_test,
+        disable_check=False
+    )
+
+    assert resource_for_test['input_metadata_file'] == upload_resource['input_metadata_file']
```
rnathuji, line 1057: Minor: Looks like the CI linter is complaining about this line being too long

bakery/src/tests/test_bakery_scripts.py
c0ldcalamari: ah, there was much confusion here, I started on this test, and got it to pass with one resource but jankily and then moved to  the other tests and forgot to come back to make the changes to include checks for two resources here, should be good now. thank you!

bakery/src/tests/test_bakery_scripts.py
c0ldcalamari: with the same regard to this, 
"ah, there was much confusion here, I started on this test, and got it to pass with one resource but jankily and then moved to the other tests and forgot to come back to make the changes to include checks for two resources here, should be good now. thank you!"

thank you!

PR #177, opened 2020-08-26
add module to debug output

ckline-tryptic:


bakery/src/scripts/link_extras.py
```diff
@@ -73,8 +73,9 @@ def match_canonical_book(canonical_ids, containing_books):
         )
     except StopIteration:
         raise Exception(
-            "Multiple containing books, no canonical match!\n"
-            + f"{containing_books}"
+            "Multiple containing books, no canonical match!\n" +
+            f"module: {module_uuid}\n" +
```
rnathuji, line 77: I tried using this branch to debug the error with `col28330`, and I actually found that providing the faulty `node.attrib['href']` or legacy ID instead of the converted module UUID was more useful as it allowed me to quickly find the link the in the CNXML and also run the extras query. Would that be better to print here for our purposes? 🤷‍♂️

tomjw64 (approved):
No harm in keeping both, right? :) Anyways, approving.

rnathuji:
Please do go ahead and merge this, but I'll reopen [cnx 1144](https://github.com/openstax/cnx/issues/1144) as I don't believe that corresponding debug output made it into this PR.

PR #181, opened 2020-08-31
improve Google Docs conversion, add greybox

therealmarv:
implementation details are here:

https://github.com/openstax/gdoc-spike/pull/1


therealmarv:
@rnathuji good catch... I will "merge" that changes. Somehow did not compared it well enough before because it was not "code"

PR #182, opened 2020-08-31
Add utility function for book slug lookups

rnathuji:
For now, this utility function relies upon a static JSON file to
lookup and return book slugs given a book UUID.


bakery/src/scripts/utils.py
```diff
@@ -40,3 +45,10 @@ def parse_uri(uri):
         return None
     legacy_id, legacy_version = uri.split('@')
     return legacy_id, legacy_version
+
+
+def book_uuid_to_slug(book_uuid):
```
tomjw64, line 50: If we plan to call this more than once during some step, do we care about doing the io here for the included file every time it is called?

rnathuji: Good point. If we end up going in this direction, I can cleanup that aspect.

rnathuji:
Closing this PR in deference to [PR 16](https://github.com/openstax/content-manager-approved-books/pull/16) in the ABL repo and changes in [PR 190](https://github.com/openstax/output-producer-service/pull/190) to pull / use slug data from there.

PR #190, opened 2020-09-15
Rewrite book links in gdocify to REX URLs

rnathuji:
This PR includes the following changes:

* An addition to `fetch-book` to download the `book-slugs.json` from our ABL repo alongside raw content data
* Minor fixes / improvements to `link-extras` found during development / testing
* Change to `link-extras` to lookup and insert page slug data alongside the canonical book UUID as a data attribute
* Updates to `gdocify-book` to rewrite both inter-book and intra-book links with REX URLs


bakery/src/scripts/gdocify_book.py
```diff
@@ -3,32 +3,108 @@
...
+            page_uuid = page_link
+            page_fragment = None
+
+        # This is either an intra-book link or inter-book link. We can
+        # differentiate the latter by data-book-uuid attrib).
+        if node.attrib.get("data-book-uuid"):
+            external_book_uuid = node.attrib["data-book-uuid"]
+            external_book_slug = book_slugs_by_uuid[external_book_uuid]
+            external_page_slug = node.attrib["data-page-slug"]
+            node.attrib["href"] = _rex_url_builder(
```
tomjw64, line 64: I wonder if this should be somewhere shared with the PDF pipelines (I assume we want REX linking behavior in that pipeline as well. It's work that's perhaps not in scope with this PR, but I thought I would bring it up.

tomjw64 (approved):

bakery/src/tasks/fetch-book.js
```diff
@@ -30,6 +31,7 @@ const task = (taskArgs) => {
           cd fetched-book
           book_dir="$(cat ../book/collection_id)"
           yes | neb get -r -d "$book_dir/raw" "$(cat ../book/server)" "$(cat ../book/collection_id)" "$(cat ../book/version)"
+          wget ${bookSlugsUrl} -O "$book_dir/book-slugs.json"
```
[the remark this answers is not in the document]
rnathuji: Yes, it will. I think we'll end up needing / using it in the preview part of COPS when we implement [cnx 1163](https://github.com/openstax/cnx/issues/1163).

bakery/src/scripts/gdocify_book.py
rnathuji: Hmmm...I'm not sure (re: plans for the link rewriting on the PDF path). If so, perhaps we can refactor this a bit at that time to incorporate a common utility function?

PR #194, opened 2020-09-17
Work around boto stubber request order enforcement

tomjw64:
We don't care about the order in which uploads occur, and there is potential for both the file system glob or the thread pool to reorder s3 uploads and checks. Botocore stubber does not support order-independent requests. This is all I could do to handle the race condition in the tests short of sending in a PR against botocore (which maybe wouldn't be a bad idea, because this is a missing feature in my eyes, and it would be better in terms of correctness)


rnathuji (approved):
Nice find + work around!

PR #199, opened 2020-09-22
Use traefik basic auth with docker swarm secret

tomjw64:
tested on staging :)

### DIFF
--- a/docs/operations/setting_up_the_swarm.rst
+++ b/docs/operations/setting_up_the_swarm.rst
@@ -203,3 +203,16 @@ Create Main Traefik Service
      --entrypoints='Name:https Address::443 TLS:/etc/ssl/certs/openstax.crt,/etc/ssl/private/openstax.pem' \
      --logLevel=INFO \
      --accessLog
+
+----
+
+***********************
+Set Up Required Secrets
+***********************
+
+The stack requires that the docker secret ``basic-auth-users`` is set in the swarm to work properly.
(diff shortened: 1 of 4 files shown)


docs/operations/setting_up_the_swarm.rst
```diff
@@ -202,4 +202,17 @@ Create Main Traefik Service
...
+***********************
+Set Up Required Secrets
+***********************
+
+The stack requires that the docker secret ``basic-auth-users`` is set in the swarm to work properly.
+An example of creating basic auth credentials with a single user is the following, when ``DOCKER_HOST`` is properly pointing to the running swarm:
+
+.. code-block:: bash
+
+  htpasswd -nbB <username> <password> | docker secret create basic-auth-users -
```
rnathuji, line 218: Questions vs code comments: If we wanted to set multiple users over time or update passwords, would we basically rotate the secret? I don't see a way to retrieve the secret value (inspect seems to just give metadata 🤷‍♂️ ), so would we create a new password file with `htpasswd` and then push up? (e.g. in lieu of an append)

rnathuji (approved):

docs/operations/setting_up_the_swarm.rst
tomjw64: That's correct. We would have to rotate (and that also means that everytime we do, we must know ALL existing credentials)

tomjw64: Maybe we can keep a secret in aws secretsmanager, just to "remind us" what the existing secret value is when we need to rotate...

PR #201, opened 2020-09-25
Fix unconverted math issues in Google Docs

rnathuji:
This PR fixes two issues which resulted in unconverted math in early Google Docs testing:

1. Broken conversions due to injection of `\mspace` when translated to LaTeX: This is fixed by updating our version of `pandoc` and underlying dependencies.
2. Broken conversions due to `<mtext>` elements with attribute `mathvariant="bold-italic"`: This is fixed by converting to `<mi>` as a workaround for `texmath` otherwise generating `LaTeX` which it causes it to error.

These changes should fix all math conversion warnings found during `convert-docx` when building Calculus Volume I (`col11964`) except for [3 errors in `m53596`](https://github.com/openstax/cnx/issues/1164#issuecomment-698604925) which should probably be addressed by content fixes instead of code patching.

### DIFF
--- a/bakery/src/scripts/Dockerfile
+++ b/bakery/src/scripts/Dockerfile
@@ -3,7 +3,7 @@ FROM python:3.7-slim
 LABEL maintainer="OpenStax Content Engineering"
 
 RUN apt update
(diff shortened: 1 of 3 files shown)


therealmarv (approved):
looks nice

PR #206, opened 2020-10-19
Miscellaneous cops clean up

c0ldcalamari:
- Update documentation for COPS deploy
- Fix WebDriverException from selenium - ref: https://openstax.slack.com/archives/C0LA54Q5C/p1603206139267400?thread_ts=1603205856.266700&cid=C0LA54Q5C
- Separate concerns for docker compose and docker swarm.

Docker Swarm does not use `depends_on`
Docker Compose was generating a stack file that was no longer compatible. 
ref: https://docs.docker.com/compose/compose-file/


docs/operations/updating_the_stack.rst
```diff
@@ -79,178 +94,134 @@ You can copy down your ``cops.pem`` into your ``~/.ssh`` from bastion2 by:
...
+0. Update Buildout and JS Dependencies
 ===========================================
 
-Keep sure you are on the latest master branch.
+Make sure you are on the latest master branch.
+
+.. code-block:: bash
+
+   $ cd output-producer-service
+   $ git checkout master
```
rnathuji, line 111: We haven't had to worry about this too because our merge velocity is fairly light, but technically (to be safe), it's best to checkout the specific git ref associated with the `output-producer-service` repo in the build of the code version tag selected for a deployment. Is it worth making this clear as part of these updates?

c0ldcalamari: yes it is, will do - thank you!

m1yag1 (approved):

PR #209, opened 2020-10-27
Add git-storage compatible tasks

tomjw64:
Despite an atrocious git history on this, changes to existing files are minimal, and so this should hopefully be fairly safe. A flow chart is added to the bakery README to assist anyone wanting to try out some of these new subcommands.


.circleci/config.yml
```diff
@@ -80,7 +80,7 @@ jobs:
       - run:
           name: Lint bakery Python scripts
           command: |
-            flake8 bakery/src/**/**.py
+            flake8 bakery/src/**/**.py --max-line-length=100
```
rnathuji, line 83: It looks like this is mostly to avoid breaking the regex string [here](https://github.com/openstax/output-producer-service/blob/e031b35c0ece56b204f9fd7bf3688a3f7b9f8213/bakery/src/scripts/link_single.py#L27) which makes sense. Maybe instead of a global change to line-length we can just add an in-line exception by appending that line with `# noqa: E501`? Or do we want to relax this linter rule more broadly anyways?

bakery/src/tasks/checksum-book.js
```diff
@@ -33,8 +33,9 @@ const task = (taskArgs) => {
           cp -r baked-book/* checksum-book
           book_dir="checksum-book/$collection_id"
           mkdir "$book_dir/baked"
+          echo test
```
rnathuji, line 36: Minor: Maybe left over from testing and can be removed?

bakery/README.md
```diff
@@ -62,6 +62,11 @@ To build the PDF:
 
 `bakery run build-pdf col30939 -d ./data/prealgebra-2e/`
 
+### New Git-Storage Compatible Tasks
+Git-storage compatible tasks are not ready for use outside of the CE Tech yet.
+The following graph represents the (almost) current state of git-storage compatible tasks. Intermediate states are shown in blue, tasks in yellow.
+![Flow for git-storage compatible pipelines](./docs/pipelines.png)
```
rnathuji, line 68: Still wrapping my head around it 😅 , but very useful visual to have in the docs!

Is it worth adding a note here on how to edit / update in the future? (e.g. I think you used yEd based on the Slack thread, but not sure)

.circleci/config.yml
tomjw64: Heh, yeah. I think that I've finally had it with the incredibly tiny max line length prescribed by E501. It would probably help if flake8 actually helped us format, but autopep8 barely helped in this case. What do you think?

bakery/README.md
tomjw64: Yeah, I'll add a note

.circleci/config.yml
rnathuji: Sure, I have no problem if we want to increase the limit in general for the codebase, just wanted to check.

bakery/src/tasks/assemble-book-group.js
```diff
@@ -0,0 +1,62 @@
...
+            mv "${rawCollectionDir}/metadata/$slug_name.metadata.json" "${rawCollectionDir}/modules/metadata.json"
+            
+            # Assembly destination must nested EXACTLY one level under cwd for symlinks to work 
+            neb assemble "${rawCollectionDir}/modules" temp-assembly/
+
+            # We shouldn't we need this symlink
+            rm temp-assembly/collection.xml
+
+            find temp-assembly -type l | xargs -I{} cp -P {} "${symlinkOutput}"
+            find "${symlinkOutput}" -type l | xargs -I{} cp -P {} "${assembledOutput}"
```
rnathuji, line 52: Question / note for myself: The why of this will probably become more obvious as I get to where they're used, but it seems we're populating the same symlinks in both outputs.

bakery/src/tasks/assemble-book-metadata.js
```diff
@@ -33,7 +33,13 @@ const task = (taskArgs) => {
           book_dir="assembled-book/$collection_id"
           target_dir="assembled-book-metadata/$collection_id"
           mkdir "$target_dir"
-          assemble-meta "$book_dir" "$target_dir/collection.assembled-metadata.json"
+
+          echo "{" > uuid-to-revised-map.json
+          find assembled-book/$collection_id/raw/ -path */m*/metadata.json | xargs cat | jq -r '. | "\"\(.id)\": \"\(.revised)\","' >> uuid-to-revised-map.json
+          echo '"dummy": "dummy"' >> uuid-to-revised-map.json
```
rnathuji, line 39: Is there a reason we're adding these `dummy` entries for `uuid-to-revised-map.json` and `module-canonicals.json` in various task scripts?

tomjw64: JSON does not allow trailing commas, so I have to include and extra entry without one. I was hoping there was a better way to create valid json, but this was the only thing I came up with in a reasonable amount of time. Know of a way that might be more neat?

rnathuji: Ah, I see. Nothing obvious comes to mind in terms of alternatives. 🤷‍♂️

bakery/src/tasks/assemble-book-group.js
tomjw64: To potentially help with understanding, there are two reasons to add symlinks:
1. A step needs to resolve links to resources. This is for correctness.
2. The symlinks will allow images to show if a developer opens a markup file containing links between steps. This is for developer convenience, not correctness.

If we dump symlinks to an input to a task, it's probably reason 1 (and you'll notice `fetched-book` is also included as an input - so the links have something to resolve to). If we dump symlinks into an output, it's probably reason 2 (because the links will actually make their way into the output directory in the file system, and it's is not necessary to include `fetched-book` as an input to the task)

rnathuji: Cool, thanks for the breakdown on that! I suspect the symlinks on this on are reason 2 as I didn't see them being used anywhere later in scripts, but it makes sense that it would be convenient to see images when you view the XHTML.

bakery/src/scripts/disassemble_single.py
```diff
@@ -0,0 +1,193 @@
...
+            )
+
+    with open(f"{out_dir}/{book_slug}.toc.xhtml", "wb") as out:
+        out.write(etree.tostring(toc, encoding="utf8", pretty_print=True))
+
+    with open(f"{out_dir}/{book_slug}.toc-metadata.json", "w") as toc_json:
+        json.dump(book_toc_metadata, toc_json)
+
+
+if __name__ == "__main__":
```
rnathuji, line 192: Minor comment: This script looks pretty similar to `disassemble_book.py`. I'm not sure if there's going to be a need to maintain these in parallel for some non-trivial period, but if so (or it's unknown) maybe the shared stuff can be refactored in some manner? 🤷‍♂️  (if we think this is worth doing, it can be a separate issue / PR, just want to raise the thought)

rnathuji:
Minor comment: It may be early just now, but perhaps worth creating an issue to follow up with unit tests for the new scripts.

bakery/src/scripts/disassemble_single.py
tomjw64: hmmm yeah... in fact, it's nearly the exact same, the only meaningful differences the inputs and outputs and preventing the using the `collection` part of the output basename. A diff:
```
$ diff ./scripts/disassemble_book.py ./scripts/disassemble_single.py
42,47c42,45
<     in_dir = Path(sys.argv[1]).resolve(strict=True)
<     out_dir = (in_dir / "disassembled").resolve(strict=True)
<     collection_uuid, collection_version = sys.argv[2:4]
<     baked_file = (in_dir / "collection.baked.xhtml").resolve(strict=True)
<     baked_metdata_file = \
<         (in_dir / "collection.baked-metadata.json").resolve(strict=True)
---
>     xhtml_file = Path(sys.argv[1]).resolve(strict=True)
>     metadata_file = Path(sys.argv[2]).resolve(strict=True)
>     book_slug = sys.argv[3]
>     out_dir = Path(sys.argv[4])
49c47
<     with open(baked_file, "rb") as file:
---
>     with open(xhtml_file, "rb") as file:
52d49
< 
55c52
<     with open(baked_metdata_file, "r") as baked_json:
---
>     with open(metadata_file, "r") as baked_json:
70c67
<         id_with_context = f'{collection_uuid}@{collection_version}:{doc.id}'
---
>         id_with_context = f'{binder.ident_hash}:{doc.id}'
188c185
<     with open(f"{out_dir}/collection.toc.xhtml", "wb") as out:
---
>     with open(f"{out_dir}/{book_slug}.toc.xhtml", "wb") as out:
191c188
<     with open(f"{out_dir}/collection.toc-metadata.json", "w") as toc_json:
---
>     with open(f"{out_dir}/{book_slug}.toc-metadata.json", "w") as toc_json:
```

Maybe there's a way to change disassemble_book to accept both tasks, kind of like we did with assemble-meta. I'll take a look.

bakery/src/tests/data/collection.assembled-metadata.json
```diff
@@ -5,7 +5,7 @@
     "m42092@1.10": {
         "abstract": "Explain the difference between a model and a theory"
     },
-    "col11406@1.17": {
```
tomjw64, line 8: @rnathuji I want to double check the change in this test data as well as in `data/collection.baked-metadata.json` are correct. iirc, we should never see `colxxxxx@v.v` anymore, correct? (because of the change you made to always include metadata in `neb get`)?

bakery/src/scripts/disassemble_single.py
tomjw64: Okay, I did my best, but it did require a change in some regression test data. Would appreciate another set of eyes on that.

bakery/src/tests/data/collection.assembled-metadata.json
rnathuji: Yeah, that's correct.

rnathuji (approved):

PR #210, opened 2020-10-27
add math table to png conversion

therealmarv:


bakery/src/scripts/mathmltable2png.py
```diff
@@ -0,0 +1,188 @@
...
+def get_png_dimensions(png_bytes):
+    check = struct.unpack('>i', png_bytes[4:8])[0]
+    if check != 0x0d0a1a0a:
+        return 0, 0
+    width, height = struct.unpack('>ii', png_bytes[16:24])
+    return width, height
+
+
+def mathml2svg_jsonrpc(equation):
+    url = "http://localhost:3000/jsonrpc"
```
rnathuji, line 84: Minor: This URL is defined twice here and in `svg2png_jsonrpc`. Maybe worth defining as a constant or refactoring a general shared function for making a request to the server?

bakery/src/scripts/Dockerfile
```diff
@@ -28,6 +28,12 @@ RUN wget https://github.com/jgm/pandoc/releases/download/${PANDOC_VERSION}/pando
...
-COPY ./*.py ./*.json /code/scripts/
+RUN curl -sL https://deb.nodesource.com/setup_14.x | bash -
+RUN apt-get install -y nodejs
+
+COPY ./*.py ./*.js ./*.json /code/scripts/
 COPY ./gdoc/ /code/gdoc/
 RUN pip install /code/scripts/.
+
+RUN npm --prefix /code/scripts install /code/scripts
+RUN npm install forever@3.0.2 -g
```
rnathuji, line 39: Minor: Would it make developer iterations faster (e.g. when doing `docker build src/scripts/.` after a code change) if we move this to just after line 32 🤷‍♂️ ?

therealmarv: I changed the order of commands in the last commit. But as you said I thing it has maybe only minor impact.

rnathuji (approved):

PR #214, opened 2020-10-29
Bn/cnx#1247/archive preview wrap up

c0ldcalamari:
Main updates were for ENV variables


tomjw64 (commented):
Do we need to rewrite links generated by build-pdf and upload-book, or will links in cops continue to work as they do now (or in the case of dist-preview, continue to not work) independent of whether or not the archive-preview cloudfront changes have been deployed?

tomjw64 (approved):
Seems like it works!

bakery/env/local.json
```diff
@@ -1,14 +1,18 @@
 {
   "ENV_NAME": "local",
   "COPS_TARGET": "http://backend/api",
-  "S3_DIST_BUCKET": "",
-  "S3_QUEUE_STATE_BUCKET": "",
-  "QUEUE_FILENAME": "",
-  "FEED_FILE_URL": "",
+  "COPS_BUCKET_AKI_SECRET_NAME": "((aws-sandbox-secret-key-id))",
+  "COPS_BUCKET_SAK_SECRET_NAME": "((aws-sandbox-secret-access-key))",
+  "COPS_ARTIFACTS_S3_BUCKET": "openstax-sandbox-cops-artifacts",
```
rnathuji, line 6: Why are we adding these to local.json?

c0ldcalamari: It made sense to me, that if we were developing locally, we would still need to these keys, wouldn't we?..:thinking:

rnathuji: For the local environment we will always [pull developer creds from environment variables](https://github.com/openstax/output-producer-service/blob/48b9bcb79399a919c14591bfe5d970b443840198/bakery/src/pipeline.js#L47-L59) since your local Concourse won't be using the secretsmanager and we don't want developers having to put their credentials into this file.

c0ldcalamari: ah, ok that makes sense - thank you

rnathuji (approved):

PR #215, opened 2020-11-04
Add log retention policy for COPS pipeline

rnathuji:
Adding a 14 day retention policy for COPS jobs (PDF and preview) as I think that seems reasonable in terms of a sliding window, and we probably don't want a strict build count in case there are bursts of activity around content development milestones where we may need Concourse log details to investigate errors.

### DIFF
--- a/bakery/src/pipelines/cops.js
+++ b/bakery/src/pipelines/cops.js
@@ -23,6 +23,7 @@ const pipeline = (env) => {
     tag: lockedTag,
     ...env.dockerCredentials
   }
+  const buildLogRetentionDays = 14
 
   // FIXME: These mappings should be in the COPS resource
   const JobType = Object.freeze({
@@ -111,6 +112,9 @@ const pipeline = (env) => {
 
   const pdfJob = {
     name: 'PDF',
+    build_log_retention: {
+      days: buildLogRetentionDays
+    },
     plan: [
       { get: 'output-producer-pdf', trigger: true, version: 'every' },
       reportToOutputProducerPdf(Status.ASSIGNED),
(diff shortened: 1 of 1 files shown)


tomjw64 (approved):
I'm a little confused by the argument for setting days only. Setting the `builds` field for `build_log_retention` doesn't put an upper bound on the number of retained logs, it specifies the number of most recent jobs that _must_ be retained. So bursts of activity wouldn't affect our ability to view recent logs.
That being said, 14 days is still probably long enough that no one will be wanting to look back for a job, even if it is one of the most recent, right?

rnathuji:
@tomjw64 - So that's what I initially thought, but [per this comment](https://github.com/openstax/cnx/issues/1218#issuecomment-721837403), I think it seems to actually set a limit where it will start trimming. Though, either I'm wrong, I'm reading their documented example wrong, or their documented example is wrong. 🤷‍♂️  LMK your thoughts on that comment, though.

Assuming setting `builds` does trim, I have no strong reason for 14 days other than I figured it's a pretty extreme worst case and keeping a few extra days of builds shouldn't break things. If another value seems better to you, though, I can change the constant.

tomjw64:
:open_mouth: concourse should not have blatantly incorrect documentation sitting around like that. It seems like really unexpected behavior considering they are very specific about how it works in the docs and repeat it multiple times. Maybe it's a bug they ought to know about? We should probably file an issue, since that kind of behavior could affect some unsuspecting people more than us. Maybe I can find the time to put together a minimal reproducible example tomorrow.

No issues with this PR, though :rocket:

rnathuji:
@tomjw64 - FWIW I think [this test case](https://github.com/concourse/concourse/blob/4cb7612c5319d87208564f4ccd2199923b810d6e/atc/gc/build_log_collector_test.go#L407) hits the root of it. I believe based on their documentation that test case should expect 0 deletions instead of 1. So at best a doc bug if so?

PR #218, opened 2020-11-05
Support git-based job types in COPS

tomjw64:
Job types are not selectable at this time, since there is no pipeline
that supports dequeuing these jobs, but once there are, the
infrastructure and validation is now there to handle them in COPS. The
radio buttons can simply be enabled at that point.

One slight restructure was done in the frontend, and that was to put
data that does not need to be reactive into beforeCreate. Seems to be
the more idiomatic way of handling "constant" data in Vue.

Part of the work for https://github.com/openstax/cnx/issues/1233

### DIFF
--- a/scripts/tests.local.sh
+++ b/scripts/tests.local.sh
@@ -15,6 +15,7 @@ docker-compose \
     -f docker-compose.shared.depends.yml \
     -f docker-compose.shared.env.yml \
     -f docker-compose.dev.build.yml \
+    -f docker-compose.dev.command.yml \
     -f docker-compose.dev.images.yml \
     -f docker-compose.dev.env.yml \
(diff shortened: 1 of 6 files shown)


scripts/tests.local.sh
```diff
@@ -15,6 +15,7 @@ docker-compose \
     -f docker-compose.shared.depends.yml \
     -f docker-compose.shared.env.yml \
     -f docker-compose.dev.build.yml \
+    -f docker-compose.dev.command.yml \
```
rnathuji, line 18: Minor cleanup item: I just happened to notice this when reviewing, but perhaps we can delete the commented L27 with this addition?

rnathuji (approved):

PR #228, opened 2020-11-16
Report failures in git pdf job - prevent failure in fetch-book-group

tomjw64:
turns out github-username was an email and broke everything. used a suggestion from @rnathuji to use only the github api access token in the github fetch link and it seems to work!

### DIFF
--- a/bakery/env/prod.json
+++ b/bakery/env/prod.json
@@ -15,7 +15,7 @@
   "GDOC_GOOGLE_FOLDER_ID": "",
   "GDOC_QUEUE_FILENAME": "",
   "GDOC_FEED_FILE_URL": "",
-  "GH_SECRET_CREDS": "((github-username)):((github-api-token))",
+  "GH_SECRET_CREDS": "((github-api-token))",
   "DOCKERHUB_USERNAME": "((ce-dockerhub-id))",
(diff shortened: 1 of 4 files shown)


bakery/env/prod.json
```diff
@@ -15,7 +15,7 @@
   "GDOC_GOOGLE_FOLDER_ID": "",
   "GDOC_QUEUE_FILENAME": "",
   "GDOC_FEED_FILE_URL": "",
-  "GH_SECRET_CREDS": "((github-username)):((github-api-token))",
+  "GH_SECRET_CREDS": "((github-api-token))",
```
rnathuji, line 18: Oh, should we update the `bakery/README.md` to not include the `username:` as well?

tomjw64: it's not wrong to, but it seems like it might be unnecessary as you noted.

tomjw64:
Also I just added build_log_retention, since it was missing. Overall this should spruce up the git pdf job enough that it can go to staging and work as expected.

rnathuji (approved):

bakery/env/prod.json
tomjw64: Just confirmed it works with personal access tokens, not just api keys (i don't know if they are different tbh)

PR #230, opened 2020-11-17
Support multiple links in pdf_url - make them prettier

tomjw64:
pdf_url can now be json. if it doesn't parse, it will be the href of a single download link

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -99,6 +99,19 @@ jobs:
       - store_test_results:
           path: /tmp/test-reports
 
+  frontend-unit:
+    docker:
+      - image: circleci/node:12-buster
+
+    steps:
+      - checkout
+      - run:
+          name: Run tests
+          command: |
+            cd ./frontend
+            npm ci
+            npm run test:unit
+
   lint-bakery:
     docker:
       - image: circleci/node:12-buster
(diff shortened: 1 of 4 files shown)


rnathuji (commented):
Just to confirm, is the idea to use this `=>` form in a string defined in the pipeline task output such as [here](https://github.com/openstax/output-producer-service/blob/b131ce32ccc9e071d881c8aff106bf9408775727/bakery/src/pipelines/cops.js#L250) which would get populated [here](https://github.com/openstax/output-producer-service/blob/b131ce32ccc9e071d881c8aff106bf9408775727/bakery/src/tasks/upload-book.js#L66)?

A couple of questions:

1. Assuming yes to the above, would it work to use a JSON parsable string instead (e.g. something like `'{"text": "foobar", "href": "http://foo.com/bar.json"}'`) where the back compat can just hook onto a failure to parse?
2. Would it be useful / possible to add a unit test somewhere to capture the convention?

tomjw64:
Huh, yeah, json seems like it would be a lot better. Guess I just got hung up on backwards compatibility a bit too much...

tomjw64:
@rnathuji any ideas for where such a unit test might reside?

rnathuji:
@tomjw64 - I see. I initially thought it might just be a test we add to `./frontend/test` somewhere, but looking at it more closely now I don't think those tests are actively maintained / even run anywhere today. LMK if it's not something easy to figure out / incorporate for this PR.

rnathuji (approved):

PR #233, opened 2020-11-19
Add REX links to distribution preview

tomjw64:


bakery/src/tasks/upload-book.js
```diff
@@ -63,8 +67,23 @@ const task = (taskArgs) => {
           aws s3 cp "$book_dir/collection.toc.json" "$toc_s3_link_json"
           aws s3 cp "$book_dir/collection.toc.xhtml" "$toc_s3_link_xhtml"
 
-          echo "$toc_s3_link_json" > upload-book/toc-s3-link-json
-          echo "$toc_s3_link_xhtml" > upload-book/toc-s3-link-xhtml
+          json_url=${cloudfrontUrlWithPrefix}/contents/$book_uuid@$book_version.json
+          xhtml_url=${cloudfrontUrlWithPrefix}/contents/$book_uuid@$book_version.xhtml
```
rnathuji, line 71: Are these used anywhere? Or perhaps just staging them for now in case we decide to add them to `content_urls` later?

bakery/src/tasks/upload-book.js
```diff
@@ -14,6 +14,10 @@ const task = (taskArgs) => {
     throw Error('distBucketPath must represent some directory-like path in s3')
   }
   const distBucketPrefix = `${distBucketPath}${codeVersion}`
+  const cloudfrontUrlWithPrefix = `${cloudfrontUrl}/${distBucketPrefix}`
```
rnathuji, line 17: Do we care that `cloudfrontUrl` may be `undefined` in the distribution pipeline? (I don't think it functionally impacts anything since the output from this task never gets used there)

rnathuji (approved):
I had a couple of minor questions, but this looks good to me.

bakery/src/tasks/upload-book.js
tomjw64: Oopsies. yeah, these aren't really necessary. If we need to find what these are later, that's what git history is for. I'll delete them.

bakery/src/tasks/upload-book.js
tomjw64: I had the same thought and had to double check that it wouldn't fail (it seems to resolve to the string 'undefined'). Clearly, in the webhosting pipelines, this value is not necessary, since there is no service to report to. It seems like a code smell that the parameter has no meaning and seems to luckily not break stuff in a webhosting pipeline. Maybe we can figure out a better way to separate the upload-book task behavior between distribution-preview and webhosting pipelines in the near future, since it's probably the hardest task to reason about right now because of that.

Maybe there's a quick refactoring I can do that could help?

rnathuji: Yeah -- good point that perhaps this is an indication some form of refactoring is due. It can be in this PR (if you see some low hanging fruit), or we can create a maintenance issue to come back and look at that.

tomjw64: Did a bit of shifting around. Overall a bit happier with the upload-book task and its params now.

rnathuji (approved):
Cool -- I like the updates to refactor upload-book 🎉

PR #242, opened 2020-11-24
single-book-pipeline

ckline-tryptic:


rnathuji:
I'll echo my [question in the parent issue](https://github.com/openstax/cnx/issues/1238#issuecomment-738182464) here: Is adding four new commands (versus option flags to four existing commands that alter the behavior of the underlying bash scripts) the best approach here? It seems like adding option flags would be possible (maybe I'm wrong, though :man_shrugging: ) and DRY-er, but perhaps there is some broader directional reason this PR takes the approach it does that I'm missing?

In any case, should we create / update docs including [this diagram](https://github.com/openstax/output-producer-service/blob/master/bakery/docs/pipelines.graphml) as part of this PR?

ckline-tryptic:
I'm working on the changes Ripal and Thomas suggested. As the CLI changes and moves from the archive workflow to the git backed workflow, it makes sense to me that we should separate the git execute and tasks from the archive execute and tasks, both to make it easier to navigate the code, but also enable allow any changes like refactoring for easier maintenance without risking code that is currently in use. I would be interested to hear what other people think.

To clarify, I am only suggesting moving git tasks into a git-execute.js, and moving git related tasks and scripts into a git-tasks dir, and a git-scripts dir respectively.

PR #250, opened 2020-12-11
Remove use of page metadata.json for git bakery tasks

rnathuji:


tomjw64 (approved):
Looks great! Stuff seems to really be coming together here!

PR #257, opened 2020-12-15
single flag

ckline-tryptic:


bakery/README.md
```diff
@@ -70,6 +70,8 @@ This graph is created with yED and the source file for it is located [here](./do
 
 ![Flow for git-storage compatible pipelines](./docs/pipelines.png)
 
+The bake-group, assemble-group, bake-meta-group and assemble-meta-group tasks can be run with the --single flag to only process a single book. If the full book group has not been run through all these tasks, this flag is mandatory for running the link-single task.
```
tomjw64, line 73: why is the flag mandatory for the link-single task?
Also what is the nature of the error that occurs when a full group has not been run through the last step?

tomjw64 (commented):
Looks good. Not sure where all the changes in test_cli came from, but I suppose it doesn't really matter. I had a couple questions about behavior and also wanted to add a reminder to rebase before merging.

bakery/src/scripts/link_single.sh
```diff
@@ -1,5 +1,8 @@
 #!/bin/bash
 
 exec 2> >(tee "$LINKED_OUTPUT/stderr" >&2)
-
-link-single "$BAKED_INPUT" "$BAKED_META_INPUT" "$(cat "${BOOK_INPUT}"/slug)" "$FETCHED_INPUT/book-slugs.json" "$LINKED_OUTPUT/$(cat "$BOOK_INPUT"/slug).linked.xhtml"
+if [[ -z "${TARGET_BOOK}" ]]; then
```
rnathuji, line 4: I was quickly testing these out, and I don't think this works currently in the manner intended. I believe to cause the `cp` short circuit this should be `-n` based on the convention of passing a value if `--single` is used.

bakery/src/cli/execute.js
```diff
@@ -388,6 +388,11 @@ const tasks = {
         yargs.positional('slug', {
           describe: 'slug of collection to work on',
           type: 'string'
+        }).option('single', {
```
rnathuji, line 391: This is super small, but I'll ask anyways 🙂  - I think all of our other options are single character with a one word alias (e.g. `-p` or `--persist`). Is there a reason we're doing `--single` / `--single-book` versus, say, `-s` / `--single` for consistency?

rnathuji (commented):
I saw Thomas is already reviewing this, but I was trying the new option out from this branch and hit what I think is a small bug on `link-single`, so commenting on that and one minor query.

bakery/README.md
ckline-tryptic: It is only mandatory if it doesn't have all the books to link

bakery/src/scripts/link_single.sh
ckline-tryptic: yes that was a mistake, thanks

bakery/src/cli/execute.js
ckline-tryptic: Can do

bakery/README.md
ckline-tryptic:
```
+ exec
++ tee linked-single/stderr
+ [[ -n '' ]]
++ cat book/slug
++ cat book/slug
+ link-single baked-book-group baked-book-metadata-group precalculus fetched-book-group/book-slugs.json linked-single/precalculus.linked.xhtml
Bad href: #
Traceback (most recent call last):
  File "/usr/local/bin/link-single", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.7/site-packages/bakery_scripts/link_single.py", line 172, in main
    output_path)
  File "/usr/local/lib/python3.7/site-packages/bakery_scripts/link_single.py", line 146, in transform_links
    canonical_book_uuid = canonical_map[target_module_uuid]
KeyError: '4aa9351c-019f-4c06-bb40-d58262ea7ec7'
failed
downloading linked-single done
Error: Subprocess failed with code 1
    at completion (/home/vagrant/repos/output-producer-service/bakery/src/cli/execute.js:31:17)
    at flyExecute (/home/vagrant/repos/output-producer-service/bakery/src/cli/execute.js:225:11)
    at async handler (/home/vagrant/repos/output-producer-service/bakery/src/cli/execute.js:578:7)
cleaning up
Stopping cli_concourse_1    ... done
Stopping cli_concourse-db_1 ... done
Stopping cli_registry_1     ... done
Error: Subprocess failed with code 1
    at completion (/home/vagrant/repos/output-producer-service/bakery/src/cli/execute.js:31:17)
    at flyExecute (/home/vagrant/repos/output-producer-service/bakery/src/cli/execute.js:225:11)
    at async handler (/home/vagrant/repos/output-producer-service/bakery/src/cli/execute.js:578:7)
```

ckline-tryptic: Thats one of a number of reasons I wanted to rename the tasks to `"*-git"`.

rnathuji: This would arguably be a hack, but just as an idea: What if instead of passing `--single` to `link-single` we decide whether to short circuit based upon the existing of multiple baked XHTML files? That way it doesn't fail in this manner, doesn't require a flag, and auto-detects the best behavior. 🤷‍♂️

rnathuji: Haha...actually, by my same logic ☝️  maybe all we really need is a `--single` option on `assemble-group` and the rest are really auto-decided / detectable after that based on the inputs. Maybe it's too magical, but as a CLI user I think that usage would be pretty nifty / convenient.

ckline-tryptic: Ha, yeah I had thought about doing that for link-single, but I hadn't even considered doing that for everything after assemble-group. My thoughts on doing it for everything is, what if someone wants to assemble all the books, then bake one at a time since baking is expensive?

ckline-tryptic: I could definitely do that for link-single, but I also worried about "hackiness".

rnathuji: Yeah, true on the point that some permutations wouldn't work with the "auto-detect" logic. This comes back to my lack of clarity on exactly what use cases / flows we're targeting to support with the CLI. 

Maybe this is something we can come back to in a different PR / issue later down the road when there's more usage data to draw insights from. I'm probably muddying the waters as far as Thomas's initial question / concern anyways.

tomjw64: Well, I think as it stands in this PR, the only book you'd be able to bake by itself is the book that the fetch was scoped to. I think until we scope the task by repo instead of a single book in the repo, there will only be hackish/magical things we can do from a UX perspective.

rnathuji (approved):
Looks like CI is passing now and this is good to merge.

PR #262, opened 2020-12-16
Implement relative same-book links

tomjw64:


bakery/src/pipelines/cops.js
```diff
@@ -244,6 +245,7 @@ const pipeline = (env) => {
       taskBakeBookMeta({ image: imageOverrides }),
       taskChecksumBook({ image: imageOverrides }),
       taskDisassembleBook({ image: imageOverrides }),
+      taskPatchDisassembledLinks({ image: imageOverrides }),
```
rnathuji, line 248: Do we need to add `taskPatchDisassembledLinksSingle` to `gitDistPreviewJob` below as well?

rnathuji (commented):
Looks good overall! I just had one question on a potential additional change needed to the COPS pipeline.

bakery/src/pipelines/cops.js
tomjw64: Yup! Forgot to add it in when I rebased!

ckline-tryptic (approved):

PR #264, opened 2020-12-17
Show worker version for each job

tomjw64:


rnathuji (approved):
Nice -- this will be a very useful addition!

PR #266, opened 2020-12-18
Fix 1322 missing images

therealmarv:
convert non RGB / non greyscale color space JPEGs into RGB


philschatz (approved):
Nice! I like the "Simulate imagemagick error" test.

c0ldcalamari:
👍

PR #271, opened 2020-12-21
Fix 1322 downsampling 4 color byte images fully to 3 color byte

therealmarv:
Downsample images with the right ICC profile to 3 color byte images


bakery/src/scripts/Dockerfile
```diff
@@ -4,7 +4,7 @@ LABEL maintainer="OpenStax Content Engineering"
 
 RUN apt-get update
 RUN apt-get install -y build-essential libicu-dev pkg-config libmagic1 mime-support wget curl xsltproc lsb-release git
-RUN apt-get install -y imagemagick icc-profiles-free
+RUN apt-get install -y imagemagick icc-profiles-free curl unzip
```
rnathuji, line 7: Minor: I just happened to notice this (and it's not directly related to this PR anyways), but is there some reason why we have two consecutive `apt-get install` steps here in the `Dockerfile`?

.circleci/config.yml
```diff
@@ -75,7 +75,7 @@ jobs:
           name: Setup test environment
           command: |
             sudo apt-get update
-            sudo apt-get install -y imagemagick icc-profiles-free
+            sudo apt-get install -y imagemagick icc-profiles-free ghostscript
```
rnathuji, line 78: Minor: Should we update this dep in the `README.md` note too?

rnathuji (approved):

bakery/src/scripts/Dockerfile
therealmarv: I was not aware about that. Just informed myself now about [Docker best practices](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run) and updated the line.

.circleci/config.yml
therealmarv: yes that's important for the manual test runs. Updated.

rnathuji (approved):

PR #279, opened 2021-01-06
Update to patched axios vulnerability

tomjw64:


therealmarv (approved):

rnathuji:
Do we need to update `yarn.lock` too?

tomjw64:
yeah. facepalm that we use both...

tomjw64:
done, yarn lockfile also updated

PR #280, opened 2021-01-06
Add job details expansion and task error reporting

tomjw64:


bakery/src/pipelines/cops.js
```diff
@@ -223,9 +233,15 @@ const pipeline = (env) => {
     on_success: reportToOutputProducerPdf(Status.SUCCEEDED, {
       pdf_url: 'artifacts/pdf_url'
     }),
-    on_failure: reportToOutputProducerPdf(Status.FAILED),
-    on_error: reportToOutputProducerPdf(Status.FAILED),
-    on_abort: reportToOutputProducerPdf(Status.FAILED)
+    on_failure: reportToOutputProducerPdf(Status.FAILED, {
```
rnathuji, line 236: I'm not sure if this is possible to address if so, but if the put to S3 fails, will the error message contents end up coming from `taskBuildPdf`?

rnathuji (approved):
Looks good, and I like the UI changes! Just had one minor clarification / for my own understanding question.

bakery/src/pipelines/cops.js
tomjw64: Oh, good thinking. I will check on that.

tomjw64: It sure does.

tomjw64: I just pushed a change that adds a task that can be used to override the common log before non-task steps. It's a bit odd, and doesn't capture the specific failure, but it can print a generic message. Might be the best we can do, since I'm not really sure of any way to actually capture the output from a put step.

rnathuji: Cool, good idea. Even if the task output details aren't conveyed at least the generic error message will give the user an accurate idea of why the job failed.

PR #282, opened 2021-01-07
single-flag execute fix

ckline-tryptic:


rnathuji (changes requested):
Two comments on this PR:

1. In general, I don't think this actually fixes the bug. I believe `singleBookFlag` will always end up being an empty string because `argv.singleBook` will be `undefined`. I would expect the fix to entail using `argv.single` instead of `argv.singleBook` versus changing the string value that gets used in `taskArgs`.

2. It seems to miss the change for `assemble-group` [here](https://github.com/openstax/output-producer-service/blob/9bfb436af304d5e1e01a3c14de7a57a4cae1657b/bakery/src/cli/execute.js#L364)

Perhaps this fix can be simplified to just be the following in all 5 cases of parsing `singleBookFlag`:

```
const singleBookFlag = argv.single
```

That will end up passing `{"singleBookFlag": true}` or `{"singleBookFlag": false}` to the tasks, which I think will work with the way they make use of it (this should be confirmed of course).

ckline-tryptic:
I see. That string to variable name got me again.

rnathuji:
I think this is still missing the fix for `assemble-group` [here](https://github.com/openstax/output-producer-service/blob/9cafec5544af987c1a5f9d363589aba4b97bb6f0/bakery/src/cli/execute.js#L364).

rnathuji (approved):

PR #288, opened 2021-01-11
Fix tasks with common-log as input vs output

rnathuji:


tomjw64:
Lol. Not sure how this happened so many times. Must have been on copy/paste autopilot.

tomjw64 (approved):

PR #290, opened 2021-01-13
Use canonical.json to override module canonical book metadata

tomjw64:
After merge, we can modify the sync script to remove the `canonical-book-uuid` metadata field from modules entirely before this step.

### DIFF
--- a/bakery/src/scripts/fetch_update_metadata.py
+++ b/bakery/src/scripts/fetch_update_metadata.py
@@ -2,14 +2,30 @@
 
 import sys
 from pathlib import Path
+from glob import glob
 from pygit2 import Repository
 from datetime import datetime, timezone
 from lxml import etree
+import json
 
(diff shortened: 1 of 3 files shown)


bakery/src/scripts/fetch_update_metadata.py
```diff
@@ -27,6 +44,20 @@ def main():
     git_repo = Path(sys.argv[1]).resolve(strict=True)
     modules_dir = Path(sys.argv[2]).resolve(strict=True)
     repo = Repository(git_repo)
+    canonical_file = Path(sys.argv[3]).resolve(strict=True)
+    collection_dir = Path(sys.argv[4]).resolve(strict=True)
```
rnathuji, line 48: The base commit for this branch is pretty old (probably why there are merge conflicts), but I think this parameter is already being passed in `master`.

rnathuji (approved):

PR #291, opened 2021-01-14
rename distribution to web-hosting

ckline-tryptic:
Grep will still find many instances of the word "distribution". They should all be unrelated.


backend/app/migrations/versions/0eee70848f87_add_seed_data_for_git_job_types.py
```diff
@@ -31,7 +31,7 @@
 def upgrade():
     utcnow = datetime.utcnow()
     server_data = [{'id': 3, 'name': 'git-pdf', 'created_at': utcnow, 'updated_at': utcnow},
-                   {'id': 4, 'name': 'git-distribution-preview', 'created_at': utcnow, 'updated_at': utcnow}]
+                   {'id': 4, 'name': 'git-web-hosting-preview', 'created_at': utcnow, 'updated_at': utcnow}]
```
tomjw64, line 34: We can't change past migrations like this. To change the name, you'd need to add a new migration.

backend/app/migrations/versions/cc7ec89e9135_added_jobs_types_table_and_foreign_key_.py
```diff
@@ -40,7 +40,7 @@ def upgrade():
     utcnow = datetime.utcnow()
 
     server_data = [{'id': 1, 'name': 'pdf', 'created_at': utcnow, 'updated_at': utcnow},
-                   {'id': 2, 'name': 'distribution-preview', 'created_at': utcnow, 'updated_at': utcnow}]
+                   {'id': 2, 'name': 'web-hosting-preview', 'created_at': utcnow, 'updated_at': utcnow}]
```
tomjw64, line 43: Same as above. We can't change past migrations like this. To change the name, you'd need to add a new migration.

bakery/env/prod.json
```diff
@@ -11,8 +11,8 @@
   "DIST_BUCKET_SAK_SECRET_NAME": "((prod-web-hosting-content-gatekeeper-secret-access-key))",
   "DIST_S3_BUCKET": "openstax-web-hosting-content-gatekeeper",
   "DIST_QUEUE_STATE_S3_BUCKET": "openstax-web-hosting-content-queue-state",
-  "DIST_ARCHIVE_QUEUE_FILENAME": "distribution-queue.json",
-  "DIST_GIT_QUEUE_FILENAME": "git-distribution-queue.json",
+  "DIST_ARCHIVE_QUEUE_FILENAME": "web-hosting-queue.json",
+  "DIST_GIT_QUEUE_FILENAME": "git-web-hosting-queue.json",
```
tomjw64, line 15: Many of these still have DIST in the name.

bakery/env/staging.json
```diff
@@ -11,8 +11,8 @@
   "DIST_BUCKET_SAK_SECRET_NAME": "((aws-sandbox-secret-access-key))",
   "DIST_S3_BUCKET": "openstax-sandbox-web-hosting-content-gatekeeper",
   "DIST_QUEUE_STATE_S3_BUCKET": "openstax-sandbox-web-hosting-content-queue-state",
-  "DIST_ARCHIVE_QUEUE_FILENAME": "distribution-queue.json",
-  "DIST_GIT_QUEUE_FILENAME": "git-distribution-queue.json",
+  "DIST_ARCHIVE_QUEUE_FILENAME": "web-hosting-queue.json",
+  "DIST_GIT_QUEUE_FILENAME": "git-web-hosting-queue.json",
```
tomjw64, line 15: Same as above. Many of these still have DIST in the name. Local env file is probably also affected.

tomjw64 (changes requested):
I see this is not in code review column yet, but here's some initial comments, just in case. Looks like a couple files were deleted, but I'm guessing they were renamed and were not git-added.

PR #292, opened 2021-01-14
Implement Job Cancellation

tomjw64:
How this works:
Every top level job (corresponding to pdf, git_pdf, etc.) is run in parallel alongside a status checker job, using `fail_fast`, meaning the entire job is interrupted if any task running in parallel fails. This status checker polls the CORGI backend, attempting to see if the status changes to aborted out from under the running task. If the status checker finds this is the case, it exits with a non-zero exit code. Because the jobs are running with `fail_fast`, concourse will also as a result interrupt the running job, even if it's in the middle of baking or something. If the status checker finds a different completed state (success or failure), then it gracefully exits with a zero exit code.

Two things to keep in mind with this change:
- This will not work if we have issues with concourse not reporting a completed/failed status to the backend when a build completes. If this seems to become an issue, we need to add an `ensure` hook on each job (or some really large timeout) that will make sure that we kick the job status out of processing to some completed state so that this parallel job can exit correctly.
- Instead of having hooks for only the pipeline job like before (which was nice, since we could be sure there were no reporting gaps), we now only manage `on_abort` and `on_error` hooks on a job level. `on_failure`, and `on_success` now need to be managed at the step level, since we now have two jobs running in parallel that need to cause different side-effects on failure.

Other alternatives and why I didn't start with them:
**Have a wrapper admin layer task around every task**
- While something like this would work nicely for something like progress reporting in the job details page, we would end up with situations where someone would try to cancel a job at the beginning of baking, in which case it would take up to an hour or more to actually stop the task running in concourse and free up resources.

**Have a wrapper admin layer injected into every task**
- Essentially the same as what is done here, but add 100% more bash magic and require that all the task images have the required dependencies.

**Have the abort button send a POST to the abort endpoint of concourse**
- Would require a new specialized resource that can record and pass the build number to each build, so that it could then be tossed back over to the CORGI backend
- Running fly in a CORGI node just for this purpose doesn't really make a whole lot of sense, so an API endpoint would be preferred, but concourse's API endpoints are not advertised as being stable. And even if we wanted to send a POST to the abort endpoint of concourse, we would need to give CORGI the authorization to do so, which would take some doing.

### DIFF
--- a/backend/app/app/api/endpoints/jobs.py
+++ b/backend/app/app/api/endpoints/jobs.py
@@ -71,5 +71,13 @@ def update_job(
     job = jobs_service.get(db_session=db, obj_id=id)
     if not job:
         raise HTTPException(status_code=404, detail="Job not found")
+    # If current job status is completed state
+    if int(job.status_id) in [4,5,6]:
+        # Pipelines will try to override an completed state with
+        # assigned, processing status if unlucky
+        # Only allow updating to completed states to prevent this
+        # Don't raise HTTPException since other fields are likely valid
+        incoming_status = int(job_in.status_id)
+        job_in.status_id = incoming_status if incoming_status in [4,5,6] else job.status_id
     job = jobs_service.update(db, job, job_in)
     return job
--- a/bakery/src/pipelines/cops.js
+++ b/bakery/src/pipelines/cops.js
@@ -89,10 +89,37 @@ const pipeline = (env) => {
             do: [taskStatusCheck({ resource: resource, apiRoot: env.COPS_TARGET, image: imageOverrides })],
             on_failure: reporter(Status.ABORTED, {
               error_message: genericAbortMessage
-            }),
+            })
           }
         ]
       }
+    }
+  }
+
+  const wrapGenericCorgiJob = (jobName, resource, step) => {
+    const report = reportToOutputProducer(resource)
+    return {
+      name: jobName,
+      build_log_retention: {
+        days: buildLogRetentionDays
+      },
+      plan: [
+        {
+          do: [
+            { get: resource, trigger: true, version: 'every' }
+          ],
+          on_failure: report(Status.FAILED, {
+            error_message: genericErrorMessage
+          })
+        },
+        runWithStatusCheck(resource, step)
+      ],
+      on_error: report(Status.FAILED, {
+        error_message: genericErrorMessage
+      }),
+      on_abort: report(Status.ABORTED, {
+        error_message: genericAbortMessage
+      })
     }
   }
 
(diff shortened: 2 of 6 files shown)


backend/app/app/api/endpoints/jobs.py
```diff
@@ -71,5 +71,13 @@ def update_job(
     job = jobs_service.get(db_session=db, obj_id=id)
     if not job:
         raise HTTPException(status_code=404, detail="Job not found")
+    # If current job status is completed state
+    if int(job.status_id) in [4,5,6]:
```
rnathuji, line 75: 🤔  - In the context of the behavior we had last week where Concourse replayed some jobs, would this end up masking that in the UI since our jobs just fire-and-forget the status updates? (e.g. the change from failed -> succeeded would eventually show up, but not the intervening state changes while the job runs)

bakery/src/pipelines/cops.js
```diff
@@ -149,19 +191,19 @@ const pipeline = (env) => {
...
-      days: buildLogRetentionDays
-    },
-    plan: [
-      { get: 'output-producer-git-pdf', trigger: true, version: 'every' },
-      reportToOutputProducerGitPdf(Status.ASSIGNED, {
+  let resource
+  let report
+
+  resource = 'output-producer-git-pdf'
+  report = reportToOutputProducer(resource)
```
rnathuji, line 198: Minor: Perhaps somewhat DRY-er if this `report` is passed as input to `wrapGenericCorgiJob` in each of these cases in lieu of L100 above (and the same can probably be done to `runWithStatusCheck` in lieu of L82). Or, at least I think that could work 🤷‍♂️ .

rnathuji (approved):
A few minor comments / questions, but this looks pretty cool overall! Thanks for writing up the [notes](https://github.com/openstax/output-producer-service/pull/292#issue-555175500) on approach / alternatives you considered as well. That was very useful!

backend/app/app/api/endpoints/jobs.py
tomjw64: That's unfortunately correct. If this behavior is too broad, we can restrict it to just the aborted status.

rnathuji: I don't really have a strong feeling either way -- this was mostly just a question for my own understanding.

bakery/src/pipelines/cops.js
tomjw64: Hmmm... an interesting suggestion.  `wrapGenericCorgiJob` will never use any reporter other than the one generated from `resource`, and it will need `resource` anyways, right? I agree it's dry-er, but it also seems weird to call a function `someFunction(x, y)` where the y=f(x) in all cases for something other than performance reasons (i.e. f(x) is expensive and should not be recalculated). I'm genuinely a bit curious about this, since I'm always interested in writing cleaner code, but I'm not actually sure what's cleaner here. What do you think?

bakery/src/scripts/status_check.sh
```diff
@@ -0,0 +1,37 @@
...
+subsequent_failures=0
+while true; do
+  curl -w "%{http_code}" -o response "$API_ROOT/jobs/${job_id}" > status_code
+  timestamp=$(date '+%H:%M:%S')
+  status_code="$(cat status_code)"
+  if [[ "$status_code" = "200" ]]; then
+    subsequent_failures=0
+    current_job_status_id=$(jq -r '.status.id' response)
+    case "$current_job_status_id" in
+      2 | 3)
```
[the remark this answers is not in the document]
tomjw64: good idea

bakery/src/pipelines/cops.js
rnathuji: Yeah -- I wrote / deleted / re-pasted the comment in my review pass as well because I debated this question myself. What probably nudged me to keep the comment in was the thought that if we (for whatever reason) ever change the signature of `reportToOutputProducer` the fact that a setup reporter gets passed through might be cleaner. This may be more of a premature line of thought, though 😅, as that might never even happen .

bakery/src/scripts/status_check.sh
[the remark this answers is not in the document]
tomjw64: done

bakery/src/pipelines/cops.js
tomjw64: I see. I'm gonna err on the side of "when we come to it" on this one, I think.

bakery/env/local.json
```diff
@@ -1,7 +1,7 @@
 {
   "ENV_NAME": "local",
   "COPS_TARGET": "http://backend/api",
-  "COPS_ARTIFACTS_S3_BUCKET": "",
+  "COPS_ARTIFACTS_S3_BUCKET": "ce-thomas-sandbox",
```
rnathuji, line 4: This looks like an inadvertent change in your recent commit.

tomjw64: Lol i was very careful adding files, then i did `commit -a` like a dunce.

tomjw64: fixed

PR #303, opened 2021-01-26
Fix 808 frontend fixes

therealmarv:


frontend/store/index.js
```diff
@@ -30,23 +30,54 @@ export const getters = {
...
-    const response = await this.$axios.$get(`/api/jobs/pages/${page}?limit=${limit}`)
-    const data = []
-    response.forEach(function (item) {
-      data.push(flattenObject(item))
-    })
-    // if (data.hasOwnProperty('collection_id')) {
-    commit('REFRESH_JOBS', data)
-    // }
+    await this.$axios.$get(`/api/jobs/pages/${page}?limit=${limit}`)
+      .then((response) => {
```
tomjw64, line 61: Seems strange to use callback syntax in an async function, unless you are explicitly trying to not await something. But this only makes a single request. Could this be written using async-await and a try-catch block?

therealmarv:
improves the error page layout and error messaging (including a corgi) on non reaching backend and 404 pages.

frontend/store/index.js
```diff
@@ -30,23 +30,54 @@ export const getters = {
...
+      })
+      .catch((error) => {
+        handleErrorAxios(error)
+      })
   },
   async getContentServers ({ commit }) {
-    const response = await this.$axios.$get('/api/content-servers')
-    commit('REFRESH_CONTENT_SERVERS', response)
+    await this.$axios.$get('/api/content-servers')
+      .then((response) => {
```
tomjw64, line 76: Echoing the above comment for this function as well.

frontend/store/index.js
therealmarv: hm, good point. It's maybe strange, or not good practice, to use await when not await something. The `then()` and `catch()` blocks are promises which are called after the `$get` but since nothing is happening after await I think I mixed it them both too much up after reading [this article](https://medium.com/better-programming/should-i-use-promises-or-async-await-126ab5c98789)

tomjw64 (approved):

PR #304, opened 2021-01-28
Fix CLI bug when using single flag related to inter-book links

ckline-tryptic:


bakery/src/scripts/link_single.py
```diff
@@ -166,18 +173,31 @@ def transform_links(
...
-    (baked_content_dir,
+    (
+        baked_content_dir,
         baked_meta_dir,
         source_book_slug,
-        output_path) = sys.argv[1:5]
+        output_path,
+    ) = sys.argv[1:5]
+
+    if len(sys.argv) > 5:
```
rnathuji, line 189: I believe even if `$TARGET_BOOK` is empty in `link_single.sh`, `sys.argv[5]` will exist, no? Or am I wrong / you found a need to do this check while testing?

bakery/src/scripts/link_single.py
```diff
@@ -166,18 +173,31 @@ def transform_links(
...
+        target_book = sys.argv[5]
+    else:
+        target_book = ""
```
rnathuji, line 192: It may be worth adding a quick unit test for the invoking of `link-single` with a non-empty `target_book` string.

bakery/src/scripts/link_single.sh
```diff
@@ -1,8 +1,4 @@
 #!/bin/bash
 exec > >(tee "${COMMON_LOG_DIR}"/log >&2) 2>&1
 
-if [[ -n "${TARGET_BOOK}" ]]; then
-    cp "${BAKED_INPUT}/${TARGET_BOOK}.baked.xhtml" "${LINKED_OUTPUT}/${TARGET_BOOK}.linked.xhtml"
-else
-    link-single "$BAKED_INPUT" "$BAKED_META_INPUT" "$(cat "${BOOK_INPUT}"/slug)" "$LINKED_OUTPUT/$(cat "$BOOK_INPUT"/slug).linked.xhtml"
-fi
+link-single "$BAKED_INPUT" "$BAKED_META_INPUT" "$(cat "${BOOK_INPUT}"/slug)" "$LINKED_OUTPUT/$(cat "$BOOK_INPUT"/slug).linked.xhtml" "'$TARGET_BOOK'"
```
rnathuji, line 4: Is the use of the single quotes around `$TARGET_BOOK` purposeful here? I think even if `$TARGET_BOOK` is empty, this will get passed as `""` to the script w/o that, no?

bakery/src/scripts/link_single.py
```diff
@@ -95,13 +95,14 @@ def _parse_tree_for_slug(tree, page_uuid):
 
 
 def patch_link(node, source_book_uuid, canonical_book_uuid,
-               canonical_book_slug, page_slug):
+               canonical_book_slug, page_slug, target_book):
```
rnathuji, line 98: It looks like you're adding this new `target_book` parameter, but it's never used. Either there is intended logic missing, or the parameter can be removed 🤷‍♂️ ?

rnathuji (commented):
I left some comments in the review in terms of code questions, but it would also be helpful if there were some context in the PR / parent issue as far as what the underlying bugs were + reasoning for these subsequent fixes. (I think I mostly inferred them but not 100% sure if everything is covered w/o that or corresponding test cases 🤷‍♂️ )

bakery/src/scripts/link_single.py
```diff
@@ -152,6 +155,10 @@ def transform_links(
     ):
         link = node.attrib["href"]
 
+        if not target_book == "":
```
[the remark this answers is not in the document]
ckline-tryptic: ha yeah

bakery/src/scripts/link_single.py
ckline-tryptic: Since it was at the end, I think the shell script was dropping it.

bakery/src/scripts/link_single.sh
ckline-tryptic: I'll double-check, but it appeared to be dropping it. I was getting not enough values to unpack.

rnathuji:
@ckline-tryptic - While reviewing the updates to this PR, I found myself confused again as to the overall context and the what / why 😅 . I spent some time reproducing the issue with a summary of thoughts captured [here](https://github.com/openstax/cnx/issues/1238#issuecomment-777757727). Can you help me understand the general approach for your proposed fix as, at least based upon that (potentially mistaken) assessment, it's not immediately apparent to me how these changes address the problem. 

A few specific questions that occurred to me when taking a pass over this iteration:

* In general, I'm failing to understand why the changes to add `--single` to `dissamble-book` (which is seemingly never used) or addition of [this fallback](https://github.com/openstax/output-producer-service/blob/74ddc9dd61210d741305e83a33276e6da423e32c/bakery/src/scripts/disassemble_book.py#L102) are present.
* Isn't [this logic](https://github.com/openstax/output-producer-service/blob/74ddc9dd61210d741305e83a33276e6da423e32c/bakery/src/scripts/link_single.py#L158-L160) going to end up rewriting both same-book and other-book links as `mock-inter-book-link`?
* Will the `target_book` passed [here](https://github.com/openstax/output-producer-service/blob/74ddc9dd61210d741305e83a33276e6da423e32c/bakery/src/scripts/link_single.py#L195) end up being the same value as `source_book_slug`?


Also, I know this issue / bug fix has been lingering for a while, so I'm happy to pair if that's useful to knock this out versus just peppering you with feedback.

rnathuji:
@ckline-tryptic - Since I was already  looking at this, I went ahead and built on your existing work to make some changes in a separate branch (you can see the net changes against master [here](https://github.com/openstax/output-producer-service/compare/49558b8...link-single-mock-extern-cr)). Other than removing the changes to `disassemble-book`, it basically uses an optional `--mock-otherbook` flag to `link-single` to determine when it can ignore a failure to lookup the canonical book for a page. Since the canonical map is generated using baked XHTML files, this should only happen in the `--single` case where only a subset of those files are present, and we can use that to trigger the `mock-inter-book-link` approach you setup (versus failing if a canonical book can't be found w/o the flag being set). Per the update to the test case, it doesn't modify any potential samebook links that match the xpath query now, though.

Take a look and LMK if those changes are helpful / make sense to you.

rnathuji:
@tomjw64 - Would you mind providing a review on the this PR? While working together with Chris on it, he mentioned that you had provided some feedback early on, and we want to make sure there's nothing we missed as far as potential gotchas or concerns with the approach taken.

bakery/src/scripts/link_single.py
```diff
@@ -153,7 +155,18 @@ def transform_links(
...
 
         target_module_uuid = get_target_uuid(link)
-        canonical_book_uuid = canonical_map[target_module_uuid]
+        canonical_book_uuid = canonical_map.get(target_module_uuid)
+
+        if ((not canonical_book_uuid == source_book_uuid) and mock_otherbook):
+            # If the canonical book UUID doesn't equal the current book (which
+            # includes if the lookup returned None) and we're mocking otherbook
+            # links, go ahead and insert the mock.
+            node.attrib["href"] = "mock-inter-book-link"
```
tomjw64, line 164: Should we add a mock `data-book-uuid` value to this as well, since that's the only thing that indicates something is an inter-book link?

rnathuji: [Yep](https://github.com/openstax/output-producer-service/pull/304/commits/36972b4e110bf4cccdc3ba2dd541595cc531e229)!

tomjw64 (approved):

PR #309, opened 2021-02-12
fix Ctrl+C so it results in non-zero exit status

philschatz:
If you press Ctrl+C while running the Bakery CLI it will result in a 0 exit status. This is incorrect & causes scripts that use the Bakery CLI to incorrectly believe the operation succeeded.

Here is a minimal example that illustrated the problem:

```js
require('tmp')      // <-- This is the culprit.

// Just sleep for 10 seconds so you can press Ctrl+C
const sleep = (ms) => new Promise((resolve) => setTimeout(() => resolve, ms))
sleep(10 * 1000)
```

Fixed in `tmp@0.2.0` (see https://github.com/raszi/node-tmp issue 233)


tomjw64 (approved):
!

PR #313, opened 2021-03-01
Added db migration for job types update

m1yag1:
* Renamed distribution to web-hosting in our job types.


m1yag1:
# Test Case 1

Test if the job type names with distribution have been updated to use web-hosting.

## GIVEN
* The CORGI stack is running via `docker-compose up`.

## WHEN
* A developer has checked the database for job_types id 2 and 4

## THEN
* The types have been renamed from distribution-* to web-hosting

# Test Case 2

Test if the job type names can be downgraded to use distribution

## GIVEN
* The test above has passed and the CORGI stack is running

## WHEN
* The `alembic downgrade -1` command has been executed

## THEN
* The job_types with web-hosting have been renamed back to distribution.

**Ensure you run `alembic upgrade head` to set the names back to what they should be.**

tomjw64 (approved):

PR #317, opened 2021-03-08
adding pytest-testrail and related adjustments in corgi ui tests

omehes:
added pytest-testrail to tests. so I can run auto-tests and define test case ids in testrail to mark tests with results automatically

### DIFF
--- a/backend/app/requirements/dev.txt
+++ b/backend/app/requirements/dev.txt
@@ -43,6 +43,7 @@ pypom==2.2.0
 pyrsistent==0.15.6
 pytest==5.3.1
 pytest-selenium==1.17.0
+pytest-testrail==2.9.0
 python-dateutil==2.8.1
 pyzmq==18.1.1
 requests==2.22.0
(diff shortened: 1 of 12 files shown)


omehes:
The circleci error is not related to my fixes: PermissionError: [Errno 13] Permission denied: '/app/report.html'

rnathuji:
I kicked the CI to see if the error mentioned above reproduces deterministically and it seems to. I then tried kicking a fresh CI run on `master` and the error doesn't reproduce there. I'm wondering if the annotation of the test with `@pytestrail.case("C624697")` could be causing this somehow (though, `test_create_new_job_button_is_displayed
` is also annotated in a similar way and it doesn't fail)? @m1yag1 - Have you seen something like this before by chance?

omehes:
@rnathuji @m1yag1 when I run these tests on my Mac, I don't get that error... 

```
-- Docs: https://docs.pytest.org/en/latest/warnings.html
----------------- generated html file: file:///app/report.html -----------------
============ 11 passed, 3 deselected, 1 warning in 73.67s (0:01:13) ============
```

rnathuji:
@omehes - I think I found the issue. I tried pushing a branch after [removing some of the new options](https://github.com/openstax/output-producer-service/commit/0566847ca2ac41b0ece4d5e49df17c74ec91d08d) added to `setup.cfg` and the error does not occur.

omehes:
@rnathuji In case you have time. I adjusted the setup.cfg file and now all tests pass :)

rnathuji (approved):

PR #319, opened 2021-03-17
Update check-feed and use of book-slugs.json for ABL schema changes

rnathuji:


bakery/src/scripts/check_feed.py
```diff
@@ -6,32 +6,106 @@
...
+        filter_function = _is_archive_entry
+        convert_function = _convert_archive_entry
+        book_id_key = ARCHIVE_BOOK_ID_KEY
+    elif feed_filter == "git":
+        filter_function = _is_git_entry
+        convert_function = _convert_git_entry
+        book_id_key = GIT_BOOK_ID_KEY
+    else:
+        # An unexpected filter value. Just return an empty list of books
+        return []
```
tomjw64, line 56: Are we sure we don't want to raise an error or fail more loudly if the `feed_filter` isn't something we expect?

rnathuji: We can [raise](https://github.com/openstax/output-producer-service/pull/319/commits/ae91d8c24773895926e578c8b0f32b3cedc61558) instead.

tomjw64 (approved):

PR #327, opened 2021-03-26
use relative time for jobs

philschatz:
It would be nice to see relatively how long ago jobs started and ended. Users can still hover over a time to get the date and time the job began/updated. I am not sure if this is actually requested by vendors; I just wanted to see how recently jobs succeeded and did not want to do the mental math.

(Note to self) Any objections to start calling this repo https://github.com/openstax/corgi ? 😉 

image

### DIFF
--- a/frontend/pages/index.vue
+++ b/frontend/pages/index.vue
@@ -217,9 +217,9 @@
             </td>
           </template>
           <template #item.created_at="{ item }">
-            <span>
-              {{ $moment.utc(item.created_at).local().format('lll') }}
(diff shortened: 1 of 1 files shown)


tomjw64 (approved):

tomjw64:
About the rename: Makes sense as long as we remember to update it here: https://github.com/openstax/concourse-pipelines/blob/master/image-autotag/pipeline.yml

therealmarv (approved):
👍

PR #334, opened 2021-04-09
Bn/cnx#1308/git pipeline regression test

c0ldcalamari:


tomjw64 (commented):
Looks great! Is there any reason not to make these two separate tests (is there any interaction between the archive and git tests)?

c0ldcalamari:
@tomjw64 they started out as two separate tests - archive/git, but the [CI was not liking](https://app.circleci.com/pipelines/github/openstax/output-producer-service/1242/workflows/f6147a70-32a1-4d16-87f1-4bb1ccdfef1e/jobs/5893) that we were doing the concourse registry twice (..?) - the errors here were not produced locally. 

I didn't know how to account for this / fix, I'd imagine this would be something to address in `execute.js` but I wasn't confident in messing with that part of the CLI for the scope of this work. 

I'd imagine something to take a look at with improvements to these tests.

tomjw64:
It looks like it's failing because the bakery containers are not cleaned up in between the tests.
Running `bakery stop -d` as a subprocess at the end of each test _should_ resolve the issue.

c0ldcalamari:
Discussed with @tomjw64 on stopping containers and separating pipelines

Tried adding to the end of each pipeline test:
```
const destroyContainers = spawn('node', [
  'src/cli/execute.js',
  'stop',
  '-d'
])
await completion(destroyContainers)
```

At the end of each (archive/ git) pipeline regression test.
but error was produced in CI, the same error was presented locally along with:

```
    error: 'uploading book' failed: Unexpected Response␊
    Status: 500 Internal Server Error␊
```
(which i thought was peculiar because, 'upload book' is not one of the task in the regression test pipelines
I tried destroying container to with only one pipeline and other errors occurred as well.

tomjw64 (approved):

PR #341, opened 2021-05-04
Don't fail fast in validate-xhtml task

rnathuji:
Process all XHTML files and then exit with nonzero code if any fail
validation instead of exiting on the first failure.


c0ldcalamari (approved):
No questions

PR #342, opened 2021-05-05
Add validate-cnxml task to pipelines with RelaxNG schema validation

rnathuji:


rnathuji:
Ask for feedback during CR: I couldn't find a way to avoid disabling shellcheck [SC2086](https://github.com/koalaman/shellcheck/wiki/Sc2086) [here](https://github.com/openstax/output-producer-service/blob/cd84607254f06d992bef8796e1e7040e417ed53a/bakery/src/scripts/validate_cnxml.sh#L23-L24) since adding quotes prevents the desired globbing from working, but I'm also not completely convinced there's not an alternative way to do this w/o squashing the linter.

c0ldcalamari (approved):
No questions, lgtm.

PR #361, opened 2021-06-08
Show missing files during checksum

philschatz:
During the checksum step some files may not be found. This is likely a content error but we do not fail on it yet.

But it is useful to know for debugging. Note there will still be entries like this from the astronomy book:

```
MISSING_FILE: /contents/810ba736-3e15-48e2-9491-cf0f401611c3
MISSING_FILE: /contents/a567ba98-3507-49d5-893c-c05cd42eefa5#fs-id1168048645820
MISSING_FILE: /contents/c4987953-c1a0-4df4-8d01-1d0df7ae228f
MISSING_FILE: /contents/bd9d6fca-fd83-4112-9379-482f40364ae7
```

Oh drat, those entries are probably the autogenerated end-of-chapter pages. If so then I see why these messages were suppressed

### DIFF
--- a/bakery/src/scripts/checksum_resource.py
+++ b/bakery/src/scripts/checksum_resource.py
@@ -59,6 +59,8 @@ def generate_checksum_resources_from_xhtml(filename, output_dir):
             create_symlink(img_filename, output_dir, sha1)
             create_json_metadata(output_dir, sha1, mime_type, s3_md5,
                                  img_basename)
+        else:
+            print('MISSING_FILE: {}'.format(img_filename))
 
     # get all a @href resources
     href_xpath = '//x:a[@href and not(starts-with(@href, "http") or ' \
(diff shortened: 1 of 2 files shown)


rnathuji (approved):
I added a quick update to a test case for this PR. Also, it looks like the corresponding git pipeline check [here](https://github.com/openstax/output-producer-service/blob/6146369d3af0565c794cc9529693a03bf1c0bdef/bakery/src/scripts/fetch_map_resources.py#L56-L60) has an existing message, so I think we're okay there as far as getting the same logging data.

rnathuji:
Oh, I just noticed the note [here](https://github.com/openstax/output-producer-service/pull/361#issue-664076354) regarding composite pages. Is this PR no longer useful in light of that because of low signal-to-noise, @philschatz, or should we log the info just in case there is something useful?

philschatz:
This PR is not useful until we remove the links to composite pages in the BakedHTML. A new card to remove the links needs to be created

PR #371, opened 2021-06-30
Bn/cnx#1548/git pdf links

c0ldcalamari:
This PR will let interbook links work and point to REX in PDF's created from archive or git.

### DIFF
--- a/bakery/src/cli/execute.js
+++ b/bakery/src/cli/execute.js
@@ -79,6 +79,7 @@ const extractLocalImageDetails = imageArg => {
 const input = (dataDir, name) => `--input=${name}=${dataDir}/${name}`
 const output = (dataDir, name) => {
   const outputDir = path.resolve(dataDir, name)
+  if (fs.existsSync(outputDir)) { fs.rmdirSync(outputDir, { recursive: true }) }
   return `--output=${name}=${outputDir}`
 }
 const COMPOSE_FILE_PATH = path.resolve(__dirname, 'docker-compose.yml')
@@ -879,7 +880,7 @@ const tasks = {
       await flyExecute([
         '-c', tmpTaskFile.name,
         `--input=book=${tmpBookDir.name}`,
-        input(dataDir, 'mathified-book'),
+        input(dataDir, 'rex-linked'),
         output(dataDir, 'artifacts')
       ], { image: argv.image, persist: argv.persist })
     }
@@ -919,7 +920,7 @@ const tasks = {
       await flyExecute([
         '-c', tmpTaskFile.name,
         `--input=book=${tmpBookDir.name}`,
-        input(dataDir, 'mathified-single'),
+        input(dataDir, 'rex-linked'),
         input(dataDir, 'group-style'),
         input(dataDir, 'fetched-book-group'),
         input(dataDir, 'resources'),
@@ -1410,13 +1411,18 @@ const tasks = {
     }
   },
   linkRex: (parentCommand) => {
-    const commandUsage = 'link-rex <identifier> <inputsource> <inputpath>'
+    const commandUsage = 'link-rex <identifier>'
     const handler = async argv => {
       const buildExec = path.resolve(BAKERY_PATH, 'build')
 
+      const re = /^(col)\\d{5}$/
+      const archive = re.test(argv.identifier)
+      const inputSrc = archive ? 'mathified-book' : 'mathified-single'
+      const contentSrc = archive ? 'archive' : 'git'
+
       const imageDetails = imageDetailsFromArgs(argv)
       const taskArgs = [`--taskargs=${JSON.stringify(
-        { ...imageDetails, ...{ inputSource: argv.inputsource, inputPath: argv.inputpath, contentSource: argv.contentsource } }
+        { ...imageDetails, ...{ inputSource: inputSrc, contentSource: contentSrc } }
       )}`]
       const taskContent = execFileSync(buildExec, ['task', 'link-rex', ...taskArgs])
       const tmpTaskFile = tmp.fileSync()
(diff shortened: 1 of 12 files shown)


c0ldcalamari:
After Code Review:
- [ ] update openstax/cops-bakery-scripts image on dockerhub
- [ ] write tests for otto

bakery/src/tasks/link-rex.js
```diff
@@ -0,0 +1,75 @@
...
+const task = (taskArgs) => {
+  const { inputSource, inputPath, contentSource: maybeContentSource } = taskArgs
+  const imageDefault = {
+    name: 'openstax/cops-bakery-scripts',
+    tag: 'trunk'
+  }
+  const imageOverrides = taskArgs != null && taskArgs.image != null ? taskArgs.image : {}
+  const imageSource = constructImageSource({ ...imageDefault, ...imageOverrides })
+  const contentSource = maybeContentSource != null ? maybeContentSource : 'archive'
+  const bookSlugsUrl = 'https://raw.githubusercontent.com/openstax/content-manager-approved-books/master/approved-book-list.json'
```
rnathuji, line 14: Can we take advantage of the fact that for git books `link-single` will populate `data-book-slug` for all other-books links per [these docs](https://openstax.atlassian.net/wiki/spaces/CE/pages/1759707137/Pipeline+Pipeline+Task+Definitions#Git-Links) (e.g. I think for `git` all the information you need to create the REX link is already in the XHTML) to avoid downloading the ABL for git books in this task?

bakery/src/tasks/link-rex.js
```diff
@@ -0,0 +1,75 @@
...
+      image_resource: {
+        type: 'docker-image',
+        source: imageSource
+      },
+      inputs: [
+        { name: 'book' },
+        { name: `${inputSource}` }
+      ],
+      outputs: [
+        { name: `${inputSource}` },
```
rnathuji, line 29: Would this mean two tasks have the same named output? I'm not sure if / how that would work. Would it be better to explicitly name this output and update `build-pdf` to consume it (similar impact to the CLI)?

bakery/src/pipelines/cops.js
```diff
@@ -223,6 +224,12 @@ const pipeline = (env) => {
       taskBakeBookMetadataGroup({ image: imageOverrides }),
       taskLinkSingle({ image: imageOverrides }),
       taskMathifySingle({ image: imageOverrides }),
+      taskLinkRex({
+        image: imageOverrides,
+        inputSource: 'mathified-single',
+        inputPath: '/*.mathified.xhtml',
```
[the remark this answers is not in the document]
c0ldcalamari: no, it would not, this is a string that get's concatenated, but now that you have brought it to my attention, it looks like I overlooked the inconsistency compared to `taskValidateHtml` - Will change it to have it more like `taskValidateHtml`

bakery/src/tasks/link-rex.js
c0ldcalamari: yes, thank you!

bakery/src/tasks/link-rex.js
c0ldcalamari: Yes, it amends the xthml docs and puts it back where it found it, so there isn't much disturbance to rest of the flow of the pipeline, was what I was going for. I don't think it's better or worse, I thought it would be less confusing this way and not have to additionally create another output. 

But, I will be updating this so it can make more sense -

bakery/src/cli/execute.js
```diff
@@ -79,7 +79,6 @@ const extractLocalImageDetails = imageArg => {
 const input = (dataDir, name) => `--input=${name}=${dataDir}/${name}`
 const output = (dataDir, name) => {
   const outputDir = path.resolve(dataDir, name)
-  if (fs.existsSync(outputDir)) { fs.rmdirSync(outputDir, { recursive: true }) }
```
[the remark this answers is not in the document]
c0ldcalamari: This was removed intentionally and related to [this change](https://github.com/openstax/output-producer-service/pull/371#pullrequestreview-697360328), but after changing the code in the related comment, I will add this back. 

Below was my thought process before this review and change to related above change:
--- 

I discovered with this line, we are unable to use the same input and output with the CLI. if the CLI detects an output with existing name it will remove it and its contents.

This behavior is not consistent with how the concourse pipeline tasks function, because an input can be used as an output in the same task. ([Reference: example 4](https://blog.concourse-ci.org/introduction-to-task-inputs-and-outputs/))

c0ldcalamari:
Test for Otto:

## Given
A pdf generated from archive that contains links to other books 
i.e. (university physics 2 book that contain links to university physics 3 book) 
## When
the inter-book link is clicked 
## Then
link will lead to an openstax.org url (REX)

## Given
A pdf generated from github repo that contains multiple books 
i.e. (university physics 2 book that contain links to university physics 3 book) 
## When
the inter-book link is clicked 
## Then
link will lead to an openstax.org url (REX)

Tip: For university physics 2, I used the link "Photons and Matter" which led to a page in university physics 3

bakery/src/tasks/link-rex.js
```diff
@@ -0,0 +1,75 @@
...
+              mkdir "$target_dir"
+              filename="collection.rex-linked.xhtml"
+              book_slugs_file="/tmp/book-slugs.json"
+              cat $abl_file | jq ".approved_books|map(.books)|flatten" > "$book_slugs_file"
+              ;;
+            git)
+              xhtmlfiles_path="${inputSource}/*.mathified.xhtml"
+              target_dir="rex-linked"
+              filename="$(cat book/slug).rex-linked.xhtml"
+              book_slugs_file="idontexistforGit"
```
rnathuji, line 56: Minor: If you want to avoid some of the checking for whether this file exists later, you could write an empty JSON like `echo "{}" > "$book_slugs_file"` so `book_slugs_by_uuid` ends up as an empty dict in the script (it would require slightly adjusting the fallback logic in `update_doc_links`, though). 🤷‍♂️

rnathuji (approved):
Looks good! I had a couple of subjective comments which you can feel free to ignore.

omehes:
testing: corgi-staging 20210708.145957
univ phys 1 links to univ phys 3 - ok (opens openstax.org page)
- unfortunately, before I could note it down, I lost the page with link to other univ phys book but it worked as expected :( 
univ phys 2 links to univ phys 3 - ok (opens openstax.org page)
- photons and matter link

PR #382, opened 2021-08-12
allow unicode characters for Collection ID

therealmarv:


omehes:
@therealmarv I checked the addition to the test. Looks good, only this seems to have an incorrect format
7F50BCE0-A915-4A68-93DE-2B018D2AD886_4_5005_c

omehes (approved):

therealmarv:
good catch. I fixed it.

PR #389, opened 2021-08-30
update cnx epub to 0.31.2

kswanson33:


rnathuji (approved):
Just to be sure it's also on your radar, the version [here](https://github.com/openstax/nebuchadnezzar/blob/0a4fbb08f19b38cc960a592891d20921e303741b/requirements/main.txt#L4) should also get updated (the `neb` version of `cnx-epub` impacts `assemble`, and the version here impacts a few of the steps in the pipeline).

☝️ is an unfortunate dependency management wart currently, but it'll get fixed with @philschatz 's single pipeline work 😅 .

philschatz (approved):

philschatz:
Thanks @rnathuji ! @kswanson33 created this PR because she ran into the exact issue you mentioned! 

She pulled [this neb PR](https://github.com/openstax/nebuchadnezzar/pull/208) into book-pipeline for testing and noticed that cnx-epub was still on an old version. :tada:

PR #399, opened 2021-09-30
Add version endpoint

m1yag1:


backend/app/tests/integration/api/endpoints/test_version.py
```diff
@@ -0,0 +1,17 @@
+import pytest
+import requests
+
+ENDPOINT = "version"
+
+
+@pytest.mark.integration
+@pytest.mark.nondestructive
+def test_version_get_request(api_url):
+    # GIVEN: An api url to the ping endpoint
```
m1yag1, line 10: Oops i need to change this to version. Copy paste error.

philschatz (approved):
Nice! The `set -e` in deploy.sh will cause the script to fail if `git` does not exist right? Just trying to avoid a silent error

PR #408, opened 2021-11-04
Updated fetch_map_resources script to copy all dir found in path

m1yag1:


philschatz:
Closing in favor of #409

PR #409, opened 2021-11-09
Copy iframes

ckline-tryptic:


philschatz (approved):

philschatz (changes requested):
It seems code coverage dropped. [See codecov](https://app.codecov.io/gh/openstax/output-producer-service/compare/409/tree/bakery/src/scripts/fetch_map_resources.py#D1L72).

It looks like we need to add a test that has iframes in it (both relative and absolute URLs)

philschatz (approved):

PR #430, opened 2022-01-20
Update book-schema-git: add schema version and style

TylerZeroMaster:
Does what it says on the tin: adds repo_schema_version, style_name, and style_href properties to book-schema-git.json in preparation for including them in Enki output.


philschatz:
This is blocked on making the book style field optional in the Book JSON

m1yag1 (approved):

PR #451, opened 2022-03-22
complete corgi ui playwright tests

omehes:
all corgi ui tests converted to playwright
some env file changes

PR #452, opened 2022-03-22
test_home in playwright + some setup changes

omehes:
one test per PR: test_home completely in playwright
some settup files changes

related PR: https://github.com/openstax/output-producer-service/pull/451 (closed)

### DIFF
--- a/backend/app/tests/ui/pages/home.py
+++ b/backend/app/tests/ui/pages/home.py
@@ -19,11 +19,154 @@ class HomeCorgi:
         self.create_new_job_button_is_visible.click()
 
     @property
-    def create_button_visible(self):
+    def create_job_modal_is_open(self):
+        return self.page.locator("job-modal")
+
+    @property
+    def create_button_is_visible(self):
         return self.page.wait_for_selector("button.create-button-start-job")
 
+    def click_create_button(self):
+        self.create_button_is_visible.click()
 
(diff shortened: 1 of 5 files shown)


backend/app/tests/ui/pages/home.py
```diff
@@ -19,11 +19,154 @@ def click_create_new_job_button(self):
...
+            "div:nth-child(2) > div:nth-child(4) > div > div > div.v-input__slot > div.v-select__slot > label"
+        )
+
+    def fill_server_field(self, value):
+        self.server_field.fill(value)
+
+    @property
+    def content_server_locator(self):
+        return self.page.locator(
+            "div:nth-child(2) > div:nth-child(4) > div > div > div.v-input__slot > div.v-select__slot > label"
```
m1yag1, line 149: With the next update to the CORGI UI we may want to work on some of these locators. I prefer the ability use `class_id` or `id` identifiers to simplify the important locators used for testing. It'll make the tests flaky any time there are changes to the UI. We're planning on having a bunch of those coming up!

m1yag1 (approved):

backend/app/tests/ui/pages/home.py
omehes: completely agree. In the meantime, I will check for better ones and replace, If I can...

PR #454, opened 2022-03-23
adding test_modal_field_errors to the test framework and deleting 2 outdated files

omehes:


omehes:
@m1yag1 Adding another converted test and also deleting 2 outdated tests (they are present under different test files - renamed)

m1yag1 (approved):

PR #464, opened 2022-03-28
adding test_modal_field_errors_disappear to test framework

omehes:


omehes:
@m1yag1 I undid the previous branch delete (deleted it by accident before merging to main) and did not realise that the previous test file test_e2e_pdf_git_jobs will be part of this commit too :) It has already been merged, not sure if it's OK to leave it here. The new test is test_modal_field_errors_disappear

m1yag1 (approved):

PR #469, opened 2022-04-06
parametrizing input data in all remaining corgi ui tests

omehes:


omehes:
@m1yag1 I realised that more of the ui tests input data can be parametrised, so I changed the remaining tests. It is 5 files but as they all have the same changes, I did them all in 1 commit.

m1yag1 (approved):

PR #471, opened 2022-04-07
removing sel/chrome/driver from tests.ci.sh/local.sh files

omehes:


omehes:
@m1yag1 I am removing some remnants of selenium chrome driver from these 2 files as we don't need those options anymore. I also made some little weirdnesses, that's why having 3 commits....

m1yag1:
yea i'm trying to do the same in my PR =).

omehes:
as per Mike's comment, closing this PR

m1yag1 (approved):

PR #472, opened 2022-04-08
Combine compose files

m1yag1:
* I've removed Concourse from the stack files because the bakery is being removed so I think it will become less important to spin up a local version of concourse. Future work will make it easier to deploy a PR and test concourse with the one we have already deployed to production.


m1yag1:
This is the follow up [PR](https://github.com/openstax/corgi/pull/473) based on this one. This PR needs to be merged first.

m1yag1:
@omehes would you be interested in reviewing this for me? I think you've had some experience using these files lately.

## Test local development 

These test cases should confirm that we can still run the development environment locally.

### GIVEN: 

The branch of this PR has been checked out

### WHEN

The developer runs:

```
docker-compose up -d
```

### Then

All the services that are expected to be running are running:

```
docker-compose ps

```

omehes (approved):
As per our discussions, looks good

PR #473, opened 2022-04-08
Add pr compose

m1yag1:
This PR focuses on enabling us to create PR environments in CORGI or our pull requests. When reviewing this PR keep in mind that there were assumptions made in the old way of testing in CI where we test using a backend-tests container that is spun up in the same docker stack. The values for the backend and frontend fixtures were hard-coded which made it difficult when we needed to test against an actual environment.


backend/app/tests/integration/conftest.py
```diff
@@ -1,6 +1,66 @@
...
+        "--tag",
+        default=os.getenv("TAG", "dev"),
+        help="The tag used on the images",
+    )
+    parser.addoption(
+        "--revision",
+        default=os.getenv("REVISION", "dev")
+    )
+
+    parser.addoption(
```
m1yag1, line 27: We don't pass in DEPLOYED_AT anymore. We validate for a valid datetime string we can parse. I need to remove this option.

m1yag1, line 27: We don't pass in DEPLOYED_AT anymore. We validate for a valid datetime string we can parse. I need to remove this option.

m1yag1: This is resolved.

m1yag1: This is resolved.

backend/app/tests/integration/api/endpoints/test_version.py
```diff
@@ -1,21 +1,26 @@
...
+from tests.utils import validate_datetime
+
+
 ENDPOINT = "version"
 
 
 @pytest.mark.integration
-@pytest.mark.nondestructive
-def test_version_get_request(api_url):
+def test_version_get_request(api_url, tag, stack_name, revision):
```
optimussam, line 11: ~~ Do we need a `@pytest.mark.parametrize( "api_url, tag, stack_name, revision", ("dev","dev","dev","20210101.1111111")],)`here? ~~
Actually  Nevermind, I've seen the `conftest.py`.

optimussam (approved):
Looks good to me.

optimussam (approved):

sparksam (approved):

PR #481, opened 2022-04-15
adding test_git_job.py, a complete e2e test to testframe

omehes:


omehes:
@m1yag1 The fail is expected as the test_git_job.py test expects real pdf/web preview link artifacts to run checks on them... this is why this PR waits for your review and discussion

omehes:
An old PR, already implemented and running. Closing

PR #482, opened 2022-05-03
Wrap DBSessionMiddleware request in try except block

m1yag1:
* We wrap the DBSessionMiddleware in try/except/finally block because we
  could throw any number of exceptions. Currently, we're seeing an
  connection error in SQLAlchemy due to unclosed connections.
* To help with the unclosed connection we've added a finally clause to
  ensure we always close the connection.

### DIFF
--- a/backend/app/app/middleware.py
+++ b/backend/app/app/middleware.py
@@ -1,5 +1,6 @@
 import logging
 
+from fastapi import Response
 from starlette.middleware.base import BaseHTTPMiddleware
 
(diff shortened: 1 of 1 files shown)


sparksam (approved):
Looks good to me.

TylerZeroMaster (approved):
This could still cause an error if `Session()` failed because reqiest.state.db could be undefined when you try to close it. Granted, I do not think that error would cause problems.

PR #483, opened 2022-05-18
Add DBSessionMiddleware unit test, Update Fastapi to 0.75.2, and update db session config

m1yag1:
Closes https://github.com/openstax/ce/issues/1906


TylerZeroMaster (approved):
I was a little confused about why the git job failed in the PR environment, but then I noticed: there are not any git jobs. There's just an archive job that was given a git repository as a collection id. 

Seems like there might be a problem with the script that fills in the fields in the UI.

Regardless, everything else looks good here.

PR #492, opened 2022-06-10
1834 Fix a few issues with first release of ABL button

TylerZeroMaster:
- [ ] openstax/ce#1834

Fixes tags not working and books missing from ABL version entry.

Removed `slug` path argument since metadata should be returned for all book slugs. 

Changed value written to clipboard: it used to create a new ABL entry, now it will create a version entry (which can be pasted into the versions array) if the ABL entry already exists.

### DIFF
--- a/backend/app/app/api/endpoints/abl.py
+++ b/backend/app/app/api/endpoints/abl.py
@@ -4,6 +4,6 @@ from fastapi import APIRouter
 
 router = APIRouter()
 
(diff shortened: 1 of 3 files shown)


ckline-tryptic (approved):

omehes:
verified on preview-git jobs in corgi staging
discussed with Tyler and Neil
some notes needed for the documentation

1C999CA6-A973-4B3D-B7F0-F310129F96AE

PR #504, opened 2022-08-04
Added vnc server and fluxbox to playwright image

m1yag1:


m1yag1:
We're not running the backend-tests container going forward after identifying issues with that approach. @omehes we don't need the VNC and browser capability b/c those should be done locally.

PR #506, opened 2022-08-08
Added backend url args for frontend container

m1yag1:


m1yag1:
This is no longer needed as @omehes addressed it in #510.

PR #531, opened 2023-02-27
make icons gray when they are not complete

philschatz:
This way folks can see which artifacts can be downloaded:

image

### DIFF
--- a/frontend/public/global.css
+++ b/frontend/public/global.css
@@ -66,3 +66,8 @@ button:focus {
   background-color: #f44336 !important;
   color: white;
 }
+
+img[data-is-complete="false"] {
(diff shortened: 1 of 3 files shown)


frontend/public/global.css
```diff
@@ -65,4 +65,9 @@ button:focus {
 .error {
   background-color: #f44336 !important;
   color: white;
+}
+
+img:not([data-is-complete="true"]) {
```
TylerZeroMaster, line 70: This selector is too general, it was making every image that did not have the data-is-complete="true" gray.

TylerZeroMaster (approved):

therealmarv (approved):

PR #533, opened 2023-02-28
2025 - Build frontend and backend during CI builds

TylerZeroMaster:
- [ ] openstax/ce#2025

Small update:
1. Build the frontend bundle before running unit tests so that we know it will build if a PR is merged.
2. Build the backend docker image so that we know it will build if a PR is merged.
3. Fix shellcheck checks. This fix would have prevented the typo I merged to main about a week ago. [SC1091](https://www.shellcheck.net/wiki/SC1090) is excluded because all the scripts that are sourced are being checked anyway.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -41,7 +41,7 @@ jobs:
 
   build-backend:
     docker:
-      - image: cimg/base:2023.02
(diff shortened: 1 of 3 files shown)


sparksam (approved):
Looks good to me.

PR #535, opened 2023-03-06
2024 - Add corgi script and leashed mode

TylerZeroMaster:
- [ ] openstax/ce#2024

Introducing leashed mode! See the README for more details. You should be able to run corgi in leashed mode in almost any environment that has docker-compose. I know it works in gitpod and codespaces, for example.


sparksam (approved):
Looks good to me , and the readme is straightforward.

PR #539, opened 2023-03-24
Remove global css

philschatz:
because it was not being cached properly.

Refs https://github.com/openstax/ce/issues/2041


TylerZeroMaster (approved):
I know you are not responsible, but if it's not too much trouble, could you replace the tabs with spaces before merging?

PR #554, opened 2023-07-05
Update dependencies

TylerZeroMaster:
Update dependencies on the frontend and backend.

I have been watching dependabot PRs pile up and, at this point, I thought it would be better to try to update all dependencies instead.

I also figured out why rollup was always complaining about broken sourcemaps: we were producing empty sourcemaps when we built the bundle for production. That should not happen anymore: now sourcemaps should only be generated when running in watch mode.


sparksam (approved):
Looks good to me.

omehes:
verified as part of m.m deploy to staging and prod - OK

PR #558, opened 2023-07-24
Make book field optional

TylerZeroMaster:
fixes openstax/ce#2019
depends on openstax/enki/pull/282

This makes the book field optional. When the book slug is not supplied, all books in the repository are built. If a book slug is supplied, only that book slug is attached to the job.


omehes:
verified in corgi staging - 20230814.162815
OK

omehes:
deployed and verified in corgi prod - OK

PR #562, opened 2023-08-16
qa test changes as per book field becoming optional

omehes:


TylerZeroMaster (approved):
How I understand these changes:
1. Add new selectors to get additional information that is now relevnt
2. Account for create job button being enabled before book is selected
3. Check urls are named with book slug name
4. Check that `all` is in the book column when a job is created without a book slug being specified.

omehes:
@TylerZeroMaster Yes to all 4 points you listed above :)

PR #564, opened 2023-08-25
new all book option qa test

omehes:


omehes:
This require book tooltip selector to be changed

omehes:
comments from Tyler:
Many formatting changes
You are collecting a list of links and fetching what the point to. Might consider checking that content-length is > 0 when doing this.
I do not think that the comparison between home.book_title_column_tooltip and home.job_type_icon_job_links_are_visible will work because the dialog is closed before this point. I think this would work better:
        home.click_job_id()

        assert home.job_id_dialog_is_visible
        # Get the job links while the dialog is still open
        job_link_text = home.job_type_icon_job_links_are_visible.inner_text().replace("\n", "")

        chrome_page_slow.keyboard.press("Escape")

        assert not home.job_id_dialog_is_visible

        assert home.book_title_column_tooltip.inner_text() == job_link_text
I also did a little looking and it seems like the job links are joined by '\n' instead of ' ' when you get them from the job dialog.

omehes:
I am closing this as per another PR with both tests reviewed. Will create a new PR for this test

PR #565, opened 2023-08-25
Verifies all book urls when book=all

omehes:


TylerZeroMaster (changes requested):
* Many formatting changes
* You are collecting a list of links and fetching what the point to. Might consider checking that `content-length` is > 0 when doing this.
* I do not think that the comparison between `home.book_title_column_tooltip` and `home.job_type_icon_job_links_are_visible` will work because the dialog is closed before this point. I think this would work better:
```python
        home.click_job_id()

        assert home.job_id_dialog_is_visible
        # Get the job links while the dialog is still open
        job_link_text = home.job_type_icon_job_links_are_visible.inner_text().replace("\n", "")

        chrome_page_slow.keyboard.press("Escape")

        assert not home.job_id_dialog_is_visible

        assert home.book_title_column_tooltip.inner_text() == job_link_text
```

omehes:
I had to do some file removal between PRs for branches all_book and bundle_all as I accidentally committed more files than I wanted... This branch/PR is checking whether the urls for bundle books are valid

TylerZeroMaster (approved):

PR #566, opened 2023-09-11
changing request - status code

omehes:


TylerZeroMaster (approved):
I am not sure this will fix the problem. I did `curl -I` and `requests.head` on the value for `bhref` that the test failed on, and I got 200 status codes in both cases. 

It might be a cause that the resource is not immediately available as soon as the job is done. I guess it is worth a try.

omehes:
@TylerZeroMaster You might be right. When I deployed to prod, all tests passed... Let's see how it will go in next deploys.

PR #567, opened 2023-09-12
Improved request and assertion code in the qa test

omehes:


TylerZeroMaster (approved):
I take it that the new test is checking that the job type icon does the same thing as the job id button when a book is not specified (i.e. open jobs dialog).

Looks good. Granted, I did not exhaustively check that the selectors will work.

omehes:
@TylerZeroMaster yes. something like that. I feel like selectors really show in the pipeline environment. locally they all pass... but who knows in concourse.

PR #589, opened 2024-02-09
Create an ERD for CORGI

TylerZeroMaster:
While working on the CORGI ABL update, it became clear that many people do not know exactly what data CORGI stores in the database. Since Enki has all of its steps displayed in graphs defined in Mermaid, I thought it made sense to do something similar for CORGI's database schema.

There's more information about the ERD generator in the README, but the short version is that the Mermaid diagram you see is automatically generated based on real fake data in the database. Generating the graph automatically based on data in the database:
1. ...verifies that the database schema matches our perception of it (eliminates bias).
2. ...makes the documentation easy to maintain: to update it, just run the script again.


philschatz (approved):
Just curious... generating an ERD diagram is not possible from just a schema or empty database?

PR #593, opened 2024-02-20
Add integration tests for CORGI

TylerZeroMaster:
Test that:
* All images build
* Services can start
* Jobs can be created
* Schema/ERD does not change unexpectedly
* Migration can downgrade and upgrade with job data existing in the database


philschatz (approved):
lgtm!

PR #605, opened 2024-05-20
Frontend abl table

TylerZeroMaster:
Add table to view ABL entries. This table could also be a good place to add support for deleting entries in the future.

Preview:

Screenshot 2024-05-20 at 9 56 17 AM

### DIFF
--- a/frontend/specs/spec-helpers.ts
+++ b/frontend/specs/spec-helpers.ts
@@ -1,6 +1,69 @@
 import * as Factory from "factory.ts";
 import type { ApprovedBookWithDate, Book, Job, User } from "../src/ts/types";
-import crypto from "crypto";
+
+// sfc32 credit: https://pracrand.sourceforge.net/
+function sfc32(a: number, b: number, c: number, d: number) {
+  return () => {
+    a |= 0;
+    b |= 0;
+    c |= 0;
+    d |= 0;
+    const t = (((a + b) | 0) + d) | 0;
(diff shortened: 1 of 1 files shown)


frontend/specs/spec-helpers.ts
```diff
@@ -65,13 +65,49 @@ export const jobFactory = Factory.Sync.makeFactory<Job>({
   worker_version: getCodeVersion(),
 });
 
+const uuidGetter = (count: number) => {
+  let uuid = crypto.randomUUID();
```
philschatz, line 69: :warning: (bikeshedding so feel free to ignore): This diff adds randomness into the tests... I've seen that used effectively for fuzz-testing but in general it seems like unit tests try to be more deterministic to reduce the number of independent variables.

But that might just be a me thing so I was curious. Also, this might be a fun discussion to have in our next [Roundtable](https://docs.google.com/document/d/1D1FNkjbCZxFNQLgh2FM_uiqn_7ifwi5r31P9MHs-FhI/edit) so I added it there

philschatz (approved):
Looks good! I just added a little prompt for future discussion with devs about testing approaches in general

frontend/specs/spec-helpers.ts
TylerZeroMaster: I replaced all randomness within test factories with [sfc32](https://github.com/bryc/code/blob/master/jshash/PRNGs.md#sfc32).

PR #606, opened 2024-05-20
Fix timezone

TylerZeroMaster:
Use custom SQLAlchemy type to ensure utc tz

Ensure stored values are converted to timezone-naive datetimes with utc offset. datetimes have always been stored in the database with an assumed utc offset.

This change makes the aforementioned assumption more explicit and adds support for converting datetimes from other timezones.

### DIFF
--- a/backend/app/app/db/schema.py
+++ b/backend/app/app/db/schema.py
@@ -15,12 +15,16 @@ class DateTimeUTC(sa.types.TypeDecorator):
     cache_ok = True
 
     def process_bind_param(self, value, dialect):
-        if isinstance(value, datetime):
-            if value.tzinfo is None or value.tzinfo.utcoffset(value) is None:
-                raise TypeError("Timezone is required")
+        # If the value does not have a timezone, it is assumed to be utc
+        if (
+            isinstance(value, datetime)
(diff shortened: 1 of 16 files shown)


backend/app/app/db/schema.py
```diff
@@ -1,23 +1,56 @@
...
+def utcnow():
+    return datetime.now(timezone.utc)
+
+
+class DateTimeUTC(sa.types.TypeDecorator):
+    impl = sa.DateTime
+    cache_ok = True
+
+    def process_bind_param(self, value, dialect):
+        # If the value does not have a timezone, it is assumed to be utc
```
philschatz, line 18: Why not migrate the data so that this condition doesn't have to remain in the codebase?

philschatz (approved):
:+1:

backend/app/app/db/schema.py
TylerZeroMaster: I will make a small change so that all datetimes that you want to use in database operations must be converted to utc. It will mean that all datetimes used to query the database will be required to be timezone aware, but that was already a soft requirement. Then the timezone value will be used to convert all datetimes to utc before the datetime is used with the database.

PR #610, opened 2024-05-22
adding assert for code version

omehes:


TylerZeroMaster (approved):
That control should only appear when the type is web view.

PR #616, opened 2024-06-06
more adjustments to locator strictness

omehes:


TylerZeroMaster (approved):
When upgrading to python 3.11, I needed to upgrade playwright because of a greenlets version problem. Did this break because of the upgrade? I thought I tried to run the UI tests, but maybe not.

PR #617, opened 2024-06-06
select latest book

omehes:


backend/app/tests/ui/test_e2e_webview_bundle_jobs_urls.py
```diff
@@ -46,13 +46,13 @@ def test_e2e_webview_bundle_jobs_urls(
         sopa = BeautifulSoup(ccont, "html.parser")
         books = sopa.select("div[class*=mdc-dialog__content]")
 
-        for book in books:
-            atags = book.find_all("a")
-            for atag in atags:
-                req = chrome_page_slow.request.get(atag["href"])
+        atags = books[0].find_all("a")
```
TylerZeroMaster, line 49: Should there be an assertion here to check that links exist? Otherwise the for loop might run on an empty array. Since the job is completed at this point, links should exist.

```python
assert len(atags) > 0
```

TylerZeroMaster (approved):
Maybe add that assertion, but other than that it looks good.

backend/app/tests/ui/test_e2e_webview_bundle_jobs_urls.py
omehes: yeah, good idea. thanks

PR #620, opened 2024-06-10
adjusting failing abort jobs

omehes:


TylerZeroMaster (approved):
Maybe I am misunderstanding the goal, but I would have thought that you would want to use `rows.extend` instead of `rows.append` so that the resulting list is composed of rows, rather than lists of rows. At the moment, I think a new list will be added to `rows` regardless of if the filter matches or not. Granted, the table should only contain rows, so the empty list scenario should not happen.

Then I also wonder why the list is necessary at all. It seems like this would be equivalent to what you have now:

```python
row_count = len(table)
for i in range(row_count):
    job_ids_ind = home.job_ids(i)

    if home.job_statuses(i) in job_stats:
        job_ids_ind.click()
        home.click_abort_button()

    else:
        pass
```

Regardless, what you have has worked in the past, I am just nitpicking.

PR #633, opened 2024-09-06
2262 missing abl versions

TylerZeroMaster:
part of openstax/ce#2262

[Only delete previous entries when a new one is added](https://github.com/openstax/corgi/commit/f614f769425ef752063a7763b20534e4706e2df1) 

Before:

When an existing entry included in the `to_keep` array was added again, all other entries for that book were deleted. Only the entry in `to_keep` would to remain.

After:

When an existing entry included in the `to_keep` is added again, update that entry while leaving other existing entries alone.


sparksam (approved):
Looks good to me.

PR #644, opened 2024-10-28
e2e pdf: changed the collection and assertion

omehes:


TylerZeroMaster (changes requested):
One change I would like to see is to `pdf_url`. At the moment, `pdf_url` is a byte string containing the raw binary form of the pdf. When the tests fail, this huge byte string is printed because we run the tests with the `--showlocals` option. This is not a problem with `io_file` or `pdf_read` because their `__repr__` only displays the class, like `<pypdf._reader.PdfReader object at 0x0000deadbeef>`.

I recommend replacing

```python
pdf_url = urlopen(r_url).read()

io_file = io.BytesIO(pdf_url)
```

with

```python
response = urlopen(r_url)

io_file = io.BytesIO(response.read())
```

That should prevent the aforementioned problem since the `__repr__` for the response is something like `<http.client.HTTPResponse object at ...>` (rather than a huge byte string).

TylerZeroMaster (approved):

PR #651, opened 2025-09-03
Core 1190 ephemeral books

TylerZeroMaster:
https://openstax.atlassian.net/browse/CORE-1190

related to: https://github.com/openstax/enki/pull/416

I also made REX the default consumer for the ABL since there is at least one thing--exercises--that does not filter by consumer and probably does not expect ephemeral books like ancillaries in the response.

This allows adding arbitrary build artifacts that are not linked to books existing directly in the repository. It can be used for books created by the pipeline among other potentially useful things in the future. 

Here's an example of build artifact link for an ephemeral book
Screenshot 2025-09-03 at 3 38 24 PM

Here's an example showing the new 'ancillaries' consumer in the ABL. As far as I know, everything that references the ABL does so with the correct query string parameters: https://corgi.ce.openstax.org/api/abl/?consumer=REX
Screenshot 2025-09-03 at 3 39 54 PM

### DIFF
--- a/backend/app/app/api/endpoints/abl.py
+++ b/backend/app/app/api/endpoints/abl.py
@@ -26,6 +26,10 @@ async def get_abl_info(
     version: Optional[str] = None,
     code_version: Optional[str] = None,
 ):
+    if consumer is None:
+        consumer = "REX"
+    elif consumer == "ANY":
+        consumer = None
     return abl_service.get_abl_info_database(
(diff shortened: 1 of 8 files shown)


P-Gill97 (approved):
Asides from the comment I left. PR seems to satisfy the requirements in the [issue](https://openstax.atlassian.net/browse/CORE-1190).

PR #656, opened 2026-07-15
Add filtering for multiple consumers in ABL

TylerZeroMaster:
https://openstax.atlassian.net/browse/CORE-2510


backend/app/tests/unit/snapshots/test_abl/test_get_abl/query_args1/query_get_abl.sql
```diff
@@ -1,3 +1,3 @@
 SELECT approved_book.book_id, approved_book.consumer_id, approved_book.code_version_id, approved_book.created_at, approved_book.updated_at 
 FROM approved_book JOIN consumer ON consumer.id = approved_book.consumer_id 
-WHERE consumer.name = :name_1
\ No newline at end of file
+WHERE consumer.name IN (__[POSTCOMPILE_name_1])
```
Dantemss, line 3: This postcompile thing caused by how it handles IN queries internally or something?

Dantemss (approved):

backend/app/tests/unit/snapshots/test_abl/test_get_abl/query_args1/query_get_abl.sql
TylerZeroMaster: Yeah, I made this snapshot testing myself, and it's kind of wonky. It relies on the `__str__`/`__repr__` of the query, which doesn't work well with set queries.

